PR #9, opened 2021-12-02
Infra mvp agents

JaafarRammal:
# Summary

Setup polymorphism (?) by extending a base agent and spawn custom agents

## Additional Information

- Base agent has some private fields of interest with getters (e.g. UUID)
- Example with 2 custom agents in Team 1 folder where each has a different `Run` implementation and content 
- On the simulation level we know that all these custom agents are just `Agent` type and we abstract the logic at this level

## Test Plan

- [ ] Initialized agents have the expected HP and floor values (ensures no nasty override in `New`)
- [ ] All agents in the simulation env are in the tower and vice-versa (ensures no base agent override since UUID is only generated and assigned in the base agent `New`)


jzzheng22:
Some of my changes include moving files around - I'll send you a diff on Slack in a bit.

pkg/agents/default/baseagent.go
```diff
@@ -6,15 +6,21 @@ import (
 
 	tower "github.com/SOMAS2021/SOMAS2021/pkg/infra/tower"
```
jzzheng22, line 7:
```suggestion
	"github.com/SOMAS2021/SOMAS2021/pkg/infra/tower"
```

You will need to change the package name in tower to `tower`. 
i.e. the first line in `tower.go` should be `package tower`, not `package Tower`.
Capital letters in package names are bad practice in Go.

jzzheng22 (changes requested):
I will look at this again in the morning but for now just this + the diff that I sent Jaafar.

pkg/infra/simulation/simulation.go
```diff
@@ -3,7 +3,7 @@ package simulation
 import (
 	"log"
 
-	baseagent "github.com/SOMAS2021/SOMAS2021/pkg/agents/default"
+	"github.com/SOMAS2021/SOMAS2021/pkg/agents"
 	agent1 "github.com/SOMAS2021/SOMAS2021/pkg/agents/team1/agent1"
 	agent2 "github.com/SOMAS2021/SOMAS2021/pkg/agents/team1/agent2"
```
jzzheng22:
```suggestion
	"github.com/SOMAS2021/SOMAS2021/pkg/agents/team1/agent1"
	"github.com/SOMAS2021/SOMAS2021/pkg/agents/team1/agent2"
```

If you can please change the first lines of `agent1.go` and `agent2.go` to:
`package agent1` and `package agent2` respectively <3
Package should be called whatever the folder name is called

jzzheng22 (changes requested):

pkg/infra/simulation/simulation.go
jzzheng22: (and it means you don't need to rename the import)

udai-arneja (approved):
Looks good.

jzzheng22 (changes requested):

jzzheng22 (approved):
LGTM. Thanks all

PR #11, opened 2021-12-08
Infra mvp tower

udai-arneja:
# Summary

Simulation of tower without messages complete. Including; tower&agent initiation, agent death and replacement, agent eating, health and floor access, platform moving etc.

## Additional Information

Large change in overall structure, including making a local copy of the abm library, having the loop iterations running in simulation (rather than abm) and the tower holding some core agent information. There is room to streamline and remove redundant structures, but for the mvp this is a stable release.

## Test Plan
Tested with various inputs and varying lengths of time/iterations. Further tests to be written and pushed, but has been tested well for stability.


Silvertrousers (approved):
Seems stable, will add message passing later this eve, its likely good enough for mvp just so other teams can get working

jzzheng22:
Why are we mixing between `float64` and `int` for food and HP? I see no reason for food to be a float?

Also, there isn't any reason to specify `64` bits, just use `float`. Also, we should be using `int` for everything anyway, I don't see any reason to use floats or `uint` either.

pkg/agents/baseagent.go
```diff
@@ -30,25 +28,32 @@ func NewBaseAgent(abm *abm.ABM, floor, hp int) (*Base, error) {
...
 func (a *Base) Floor() int {
-	return a.floor
+	return a.tower.GetFloor(a.id)
 }
 
 func (a *Base) ID() string {
 	return a.id
 }
+
+func (a *Base) IsDead() bool {
```
JaafarRammal, line 53: This is a big issue, overrides can basically be done implicitly on the original `IsDead` function and return wrong states. We should find a way to make these functions final or non-writeable

image

pkg/utils/abm/abm.go
```diff
@@ -0,0 +1,148 @@
...
+			} else {
+				a.agents[j].Run()
+			}
+			wg.Done()
+		}(&wg, i, j)
+	}
+
+	wg.Wait()
+	sort.Ints(agentsToRemove)
+	for index := len(agentsToRemove) - 1; index > -1; index-- {
```
JaafarRammal, line 115: If I inject a false ID I can kill a different agent on the next iteration given how the function is written 👀

JaafarRammal:
> Why are we mixing between `float64` and `int` for food and HP? I see no reason for food to be a float?
> 
> Also, there isn't any reason to specify `64` bits, just use `float`. Also, we should be using `int` for everything anyway, I don't see any reason to use floats or `uint` either.

These can be changed later, I am assuming it is temporary design decisions to allow flexibility for agents teams for the current design

pkg/infra/tower/tower.go
```diff
@@ -3,42 +3,72 @@ package tower
...
-	FoodOnPlatform  float64
-	FloorOfPlatform uint64
+	currPlatFood    float64
+	maxPlatFood     float64
+	currPlatFloor   uint64
 	mx              sync.RWMutex
-	agents          []abm.Agent
-	AgentCount      int
+	agentCount      int
+	agents          map[string]BaseAgentCore
```
jzzheng22, line 14: I think if you change this to a map of pointers to structs then you can make `setHP()` a lot cleaner i.e. it would become:
```
tower.agents[id].hp = newHP
```

jzzheng22: As well as most of your other setter functions in tower-agent.go I think

pkg/infra/tower/tower-functions.go
```diff
@@ -0,0 +1,96 @@
...
+		t.mx.RLock()
+		currHP := t.agents[id].hp
+		aType := t.agents[id].agentType
+		t.mx.RUnlock()
+		// currHP++
+		// aType++
+		t.setFloor(id, currHP, newFloor+1, aType)
+		remainingVacanies[newFloor]--
+	}
+}
```
jzzheng22, line 50:
```suggestion
func (t *Tower) reshuffle(numOfFloors int) {
	remainingVacancies := make([]int, numOfFloors)
	log.Printf("Reshuffling alive agents...")

	for i := 0; i < numOfFloors; i++ { // adding a max to each floor
		remainingVacancies[i] = t.agentsPerFloor
	}

	// allocating agents to floors randomly
	// iterate through the uuid strings of each agent
	for id := range t.agents {
		newFloor := rand.Intn(numOfFloors)

		for remainingVacancies[newFloor] == 0 {
			newFloor = rand.Intn(numOfFloors)
		}
		t.mx.RLock()
		currHP := t.agents[id].hp
		aType := t.agents[id].agentType
		t.mx.RUnlock()
		// currHP++
		// aType++
		t.setFloor(id, currHP, newFloor+1, aType)
		remainingVacancies[newFloor]--
	}
}
```

pkg/infra/tower/tower-functions.go
```diff
@@ -0,0 +1,96 @@
...
+	log.Printf("Killing agent %s", id)
+	t.mx.RLock()
+	deadAgentFloor := t.agents[id].floor
+	agentType := t.agents[id].agentType
+	t.mx.RUnlock()
+	t.missingAgents[deadAgentFloor] = append(t.missingAgents[deadAgentFloor], agentType)
+	t.mx.Lock()
+	delete(t.agents, id)
+	t.mx.Unlock()
+}
```
jzzheng22, line 23: Why do you need a list of `missingAgents`? Surely an easier way of doing this is just in `simulation.go` `simulationLoop()` you check for which agents have `hp <= 0` and replace those agents??? That way you don't need to keep track of `missingAgents`? This just seems to me like you've got a lot of extra state you have to maintain and keep consistent?

jzzheng22: This would change how `hpDecay()` is written slightly... but I'm also not convinced `hpDecay()` should be concerned about whether an agent dies or not - it should really just be responsible for decaying HP and nothing else.

pkg/infra/tower/tower-functions.go
```diff
@@ -0,0 +1,96 @@
...
+	floor := t.agents[id].floor
+	aType := t.agents[id].agentType
+	t.mx.RUnlock()
+	t.setHP(id, int(newHP), floor, aType)
+}
+
+func (t *Tower) FoodRequest(id string, foodRequested float64) float64 {
+	t.mx.RLock()
+	defer t.mx.RUnlock()
+	if t.agents[id].floor == int(t.currPlatFloor) {
```
jzzheng22, line 82:
```suggestion
	if t.agents[id].floor == t.currPlatFloor {
```

pkg/utils/abm/abm.go
```diff
@@ -0,0 +1,148 @@
...
+	"log"
+	"sort"
+	"sync"
+)
+
+type ABM struct {
+	mx     sync.RWMutex
+	agents []Agent
+
+	i     int // current iteration
```
jzzheng22, line 13:
```suggestion
	currIteration     int
```

pkg/utils/abm/abm.go
```diff
@@ -0,0 +1,148 @@
...
+
+func (a *ABM) SetReportFunc(fn func(*ABM)) {
+	a.reportFunc = fn
+}
+
+func (a *ABM) AddAgent(agent Agent) {
+	a.mx.Lock()
+	a.agents = append(a.agents, agent)
+	a.mx.Unlock()
+}
```
jzzheng22, line 46:
```suggestion
func (a *ABM) AddAgent(agent Agent) {
	a.mx.Lock()
	defer a.mx.Unlock()
	a.agents = append(a.agents, agent)
}
```

jzzheng22 (changes requested):
A lot of changes:

- [x] Fix all of the places I've commented
- [x] Change floats to ints where possible
- [x] If you absolutely insist on using floats, make sure they're of type `float` and not `float64`
- [x] Same goes with `uint`: these should all be `int`. Do not specify their sizes either.
- [x] A lot of abm functions seem to be defined but not used... but I suppose they can stay for now.
- [x] Not sure how many of the mutexes are actually necessary... but I suppose they can also stay for now.

pkg/utils/abm/abm.go
jzzheng22: Irrelevant for MVP I think: focus on good design and working simulation rather than safety/legality of agent actions

pkg/agents/baseagent.go
jzzheng22: Irrelevant for MVP I think: focus on good design and working simulation rather than safety/legality of agent actions

jzzheng22:
I feel like a lot of these issues have come from the architectural design - tower shouldn't have this much control over an agent, and agent shouldn't really need to have the tower as part of its struct.

If anything, a lot of these functions arguably belong to the simulation environment which should control the agent and the tower, rather than the tower controlling the agent but then the agent also affecting the tower.

JaafarRammal:
> I feel like a lot of these issues have come from the architectural design - tower shouldn't have this much control over an agent, and agent shouldn't really need to have the tower as part of its struct.
> 
> If anything, a lot of these functions arguably belong to the simulation environment which should control the agent and the tower, rather than the tower controlling the agent but then the agent also affecting the tower.
> 
> I feel like agents should be able to call certain tower functions such as `RequestFood()`, but the setting and decay of HP etc. should be done by the simulation environment, rather than the tower.

I do agree that some architectural decision can be made much better in terms of abstraction and state control. Yet bear in mind none of us has done GoLang before, and more than half the time was spent on understanding how the hell implementing polymorphism, virtual functions, overrides, finals, or just inheritance would work safely. All of that, and with a deadline in mind to deliver as quickly as possible to other teams something that works so they can start experimenting on their agents.

The current PR is **an MVP** and not a final deliverable. This means it is subject to changes (some can be major), but at least it works and unblocks the agent teams to create simulations and start experimenting. All the architectural **changes can be done without distributing the agent teams** given the current abstraction (they have a run function and a few getters which can be modified from base if we need to make changes)

> a lot of these functions arguably belong to the simulation environment

Yes and they can be moved later, it won't disrupt the agent teams work

> setting and decay of HP etc. should be done by the simulation environment, rather than the tower

Yes and that can be done later

pkg/agents/baseagent.go
JaafarRammal: It will make it harder to review code since we have to check people don't override by mistake. Will leave it away for now though since we need to get a working MVP out

pkg/utils/abm/abm.go
JaafarRammal: Nothing to do with safety / legality, it's a bug in the function

pkg/infra/simulation/simulation.go
```diff
@@ -7,23 +7,26 @@ import (
...
 
 type SimEnv struct {
-	FoodOnPlatform float64
-	AgentCount     []int
-	AgentHP        int
-	Iterations     int
+	FoodOnPlatform  float64
+	AgentCount      []int
+	AgentHP         int
+	Iterations      int
```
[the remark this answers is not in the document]
JaafarRammal: Only main accesses simEnv

pkg/agents/baseagent.go
JaafarRammal: Added TODO

pkg/utils/abm/abm.go
JaafarRammal: Added TODO

pkg/infra/tower/tower-functions.go
JaafarRammal: 1- Agreed. Future version we shall move states to simulation
2- Eventually `hpDecay()` would be parametrized, we'd like agent teams to PR proposals for it. We check in the function for death directly as otherwise we'd be checking right after it in a different loop without any agent state. Check can be moved to a different loop without issues

pkg/infra/simulation/simulation.go
```diff
@@ -74,149 +74,21 @@ func sum(inputList []int) int {
...
-	//move platform phase
-	//reset platform
-	//kill agents (check hunger, kill if hunger too high)
-	//replace - parameter (when they die, when they shuffle)
-	//random shuffle on Nth day - parameter
-
-	//end of sim
-	//- write to json file (could just dump the entire program state)
+	a.AddAgent(custagent)
+	tower.SetAgent(i, sE.AgentHP-3*floor, floor, uuid) // TODO: edit this call
```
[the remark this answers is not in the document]
JaafarRammal: probably random HP do experiment on agents dying and testing, all agents should have same starting HP. replaced

pkg/infra/tower/tower.go
```diff
@@ -3,42 +3,72 @@ package tower
...
-	log.Printf("A log from the tower!")
+
+	//logs
+	log.Printf("A log from the tower! Tick no: %d", t.tickCounter)
+	log.Printf("The food left on the platform = %f", t.currPlatFood)
+
+	//useful parameters
+	day := 24 * 60
+	numOfFloors := t.agentCount / int(t.agentsPerFloor)
+	platformMovePeriod := day / numOfFloors // can add min/max
```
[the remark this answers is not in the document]
JaafarRammal: can be modified as wished in another PR

pkg/infra/tower/tower.go
JaafarRammal: TODOed

JaafarRammal (approved):
Changes can be made to architecture, but it works and unblocks everyone and has all features required for MVP setup, so LGTM

jzzheng22:
Thanks for all the changes, they look good. Holding off on approval until we look at https://github.com/SOMAS2021/SOMAS2021/tree/infra-mvp-rework and see if message passing can be integrated into that quickly.

JaafarRammal (approved):

jzzheng22 (approved):
LGTM, thanks all.

JaafarRammal:
> Thanks for all the changes, they look good. Holding off on approval until we look at https://github.com/SOMAS2021/SOMAS2021/tree/infra-mvp-rework and see if message passing can be integrated into that quickly.

Happy to have a look at that branch as refactoring separately from this PR as it skips on some design decisions and access control

Issue #32, opened 2021-12-10
Add getter function `CurrPlatFood()` to tower


jzzheng22:
Enable agents to know how much food there is on the platform

PR #37, opened 2021-12-10
Added getter function CurrPlatFood

priyachh:
If agent and platform are on the same floor **return the amount of food on the platform**, otherwise return 0 (since agents should not know how much food is on the platform if it's not on their floor). 

# Summary

This PR is aiming to solve the issue #32  (Add getter function CurrPlatFood() to tower). 

## Test Plan

- Have tested it with the existing agents and it works :)

### DIFF
--- a/pkg/infra/baseagent.go
+++ b/pkg/infra/baseagent.go
@@ -54,7 +54,7 @@ func (a *Base) CurrPlatFood() float64 {
 	if platformFloor == a.floor { // only show the food on the platform if the platform is on the same floor as the agent
 		return foodOnPlatform // if agent and platform on same floor, return the amount of food on the platform
 	}
-	return 0.0 // otherwise return nothing because you should not be able to see food on the platform if its not on your floor.
+	return -1.0 // otherwise return -1 to indicate the platform is not on the floor. Agents cannot see the food on the platform if it is not on their floor
 }
 
 func (a *Base) Floor() int {
--- a/pkg/infra/baseagent.go
+++ b/pkg/infra/baseagent.go
@@ -51,7 +51,7 @@ func (a *Base) HP() int {
 
 func (a *Base) CurrPlatFood() float64 {
 	foodOnPlatform, platformFloor := a.tower.GetCurrPlatFood()
-	if platformFloor == a.floor { // only show the food on the platform if the platform is on the same floor as the agent
+       if platformFloor == a.floor || platformFloor == a.floor + 1  {// only show the food on the platform if the platform is on the same floor as the agent or directly below
 		return foodOnPlatform // if agent and platform on same floor, return the amount of food on the platform
 	}
 	return -1.0 // otherwise return -1 to indicate the platform is not on the floor. Agents cannot see the food on the platform if it is not on their floor
--- a/pkg/infra/baseagent.go
+++ b/pkg/infra/baseagent.go
@@ -49,12 +49,15 @@ func (a *Base) HP() int {
 	return a.hp
 }
 
+// only show the food on the platform if the platform is on the
+// same floor as the agent or directly below
 func (a *Base) CurrPlatFood() float64 {
-	foodOnPlatform, platformFloor := a.tower.GetCurrPlatFood()
-       if platformFloor == a.floor || platformFloor == a.floor + 1  {// only show the food on the platform if the platform is on the same floor as the agent or directly below
-		return foodOnPlatform // if agent and platform on same floor, return the amount of food on the platform
+	foodOnPlatform := a.tower.currPlatFood
+	platformFloor := a.tower.currPlatFloor
+	if platformFloor == a.floor || platformFloor == a.floor+1 {
+		return foodOnPlatform
 	}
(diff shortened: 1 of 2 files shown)


pkg/infra/baseagent.go
```diff
@@ -49,6 +49,14 @@ func (a *Base) HP() int {
 	return a.hp
 }
 
+func (a *Base) CurrPlatFood() float64 {
+	foodOnPlatform, platformFloor := a.tower.GetCurrPlatFood()
+	if platformFloor == a.floor { // only show the food on the platform if the platform is on the same floor as the agent
+		return foodOnPlatform // if agent and platform on same floor, return the amount of food on the platform
+	}
+	return 0.0 // otherwise return nothing because you should not be able to see food on the platform if its not on your floor.
```
JaafarRammal, line 57:
```suggestion
	return -1.0 // otherwise return -1 to indicate the platform is not on the floor. Agents cannot see the food on the platform if it is not on their floor
```

pkg/infra/tower.go
```diff
@@ -122,3 +122,8 @@ func (t *Tower) ResetTower() {
 	t.currPlatFood = t.maxPlatFood
 	t.currPlatFloor = 1
 }
+
+func (t *Tower) GetCurrPlatFood() (float64, int) {
```
JaafarRammal, line 126: - I think these should be two different functions for clarity?

JaafarRammal:
```suggestion
func (t *Tower) CurrPlatFood() (float64, int) {
```
In GoLang the convention for getters is straight the variable without the `Get`

pkg/infra/baseagent.go
```diff
@@ -49,6 +49,14 @@ func (a *Base) HP() int {
 	return a.hp
 }
 
+func (a *Base) CurrPlatFood() float64 {
+	foodOnPlatform, platformFloor := a.tower.GetCurrPlatFood()
```
JaafarRammal, line 53:
```suggestion
	foodOnPlatform, platformFloor := a.tower.CurrPlatFood()
```

JaafarRammal (changes requested):

pkg/infra/baseagent.go
JaafarRammal: If we return 0 for when the platform is not on the floor, then the agent cannot know if 0 means no food or no platform

JaafarRammal: We can still return 0 and have a separate Boolean function to query if the platform is on the floor or not

jzzheng22 (changes requested):
Is there a reason why `baseagent` needs `CurrPlatFood()`? Why not just call it directly from the tower?

I'd also just move the comments to directly above the function declarations:
```
// Returns the amount of food on the platform if the platform is on the same floor as the agent
// Returns -1 is the platform is not on the same floor as the agent
func (t *Tower) CurrPlatFood(agentID string) float64 {
```

pkg/infra/baseagent.go
```diff
@@ -49,6 +49,14 @@ func (a *Base) HP() int {
...
+	if platformFloor == a.floor { // only show the food on the platform if the platform is on the same floor as the agent
```
MattScottEEE, line 54: Or a.floor-1 (possibly):

Since we can communicate with the agent below, I assume we can see the food at this point too. At the same time, I can't see how this is immediately useful (potentially opinion formation regarding lying)

MattScottEEE (commented):
Changes look fine, just small points of consideration regarding line of sight?

pkg/infra/tower.go
priyachh: I've removed that function altogether as it was not needed (as per Jason's comment below), and I've changed the function name!

pkg/infra/baseagent.go
priyachh: I've bypassed putting functions in tower as Jason suggested below so this is no longer an issue.

priyachh:
I've made changes accounting for everyone's comments, thank you!

udai-arneja (approved):
Looks good, and there is space for changes regarding visibility later.

jzzheng22 (dismissed):
Amazing, thanks Priya :)

jzzheng22 (commented):
Actually... shouldn't this belong in `tower.go` rather than `baseagent.go`?

Since the amount of food on the platform is a property belonging to the tower, the getter function for that should also belong to the tower. It shouldn't be the agent's responsibility

priyachh:
> Actually... shouldn't this belong in `tower.go` rather than `baseagent.go`?
> 
> Since the amount of food on the platform is a property belonging to the tower, the getter function for that should also belong to the tower. It shouldn't be the agent's responsibility

The custagent only has access to baseagent so it needs to be in baseagent.go I believe. I think we also need to know what floor the agent is on so thats why baseagent is needed

JaafarRammal (approved):
LGTM thanks Priya

jzzheng22 (approved):

PR #45, opened 2021-12-12
Replace functions by state access in tower

JaafarRammal:
# Summary

Tower used base agent functions to access state. These functions could be overridden by custom agents accidentally.


udai-arneja (approved):
LGTM, nice

jzzheng22 (commented):
We're still keeping the getter functions in agent though right? If so, then happy to approve.

JaafarRammal:
Yes @jzzheng22 getters should only be used by custom agents otherwise they can be overridden

jzzheng22 (approved):
LGTM, thanks Jaafar

PR #46, opened 2021-12-12
Add run to CI/CD

JaafarRammal:
# Summary

Run the main code in CI/CD to make sure it executes

## Tests

https://github.com/SOMAS2021/SOMAS2021/actions/runs/1569363568

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -58,18 +58,18 @@ jobs:
       - name: Build
         run: make build
     
-    run:
-      name: Run
-      runs-on: ubuntu-latest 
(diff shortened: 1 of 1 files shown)


JaafarRammal:
image

Logs too expensive to store. Will store with frontend instead where we can overwrite old logs or something similar

jzzheng22 (approved):

jzzheng22:
Thanks Jaafar.

I think the logging issue is more that we should probably utilise more logging levels rather than just info, but also determine what things we actually want to log and what should be put as a print statement

lhl2617:
oh fuck yes

PR #47, opened 2021-12-12
Uniformity to timing parameters

udai-arneja:
# Summary

Initialising timing parameters in main.go and using these across simulation and tower. Combined the different tick counters to a common counter in the new struct.

Open to comments on streamlining this, if possible.

I think doing this for multiple things, such as agent initialisation would be useful for code readability. Passing too many parameters into functions (in particular instantiation functions) is a little difficult to read.

## Additional Information

Created a new struct in utils for overall timing parameters across the simulation and tower. Cleaner initialisation and uniformity of quantities such as ticks per day etc.

TODO: integrate into abm, but should be done after abm has been restructured.

## Test Plan

Used logs to check coherency of tick counter between simulation and tower. General checking of timing through logs - all behaving as expected.


pkg/infra/tower.go
```diff
@@ -4,21 +4,20 @@ import (
 	"math/rand"
 
 	"github.com/SOMAS2021/SOMAS2021/pkg/messages"
+	. "github.com/SOMAS2021/SOMAS2021/pkg/utils/globalTypes/Day"
```
jzzheng22, line 7: Same comment as for import in main.go

pkg/simulation/simulation.go
```diff
@@ -6,19 +6,20 @@ import (
 	"github.com/SOMAS2021/SOMAS2021/pkg/agents/team6"
 	"github.com/SOMAS2021/SOMAS2021/pkg/infra"
 	"github.com/SOMAS2021/SOMAS2021/pkg/utils/abm"
+	. "github.com/SOMAS2021/SOMAS2021/pkg/utils/globalTypes/Day"
```
jzzheng22, line 9: Same comment as for import in main.go

pkg/utils/globalTypes/Day/day.go
```diff
@@ -0,0 +1,21 @@
+package day_type
+
+type DayInformation struct {
+	IterationsPerDay       int
+	SimulationDays         int
+	DaysPerReshuffle       int
+	IterationsPerReshuffle int
+	TotalIterations        int
+	CurrTick               int
```
jzzheng22, line 9:
```suggestion
	TicksPerDay       int
	SimulationDays         int
	DaysPerReshuffle       int
	TicksPerReshuffle int
	TotalTicks        int
	CurrTick               int
```

jzzheng22: Swapping between iterations and ticks gets confusing - better to just call them Ticks I think

jzzheng22: Obviously will need to change this wherever iteration has been used

pkg/simulation/simulation.go
```diff
@@ -59,22 +59,22 @@ func (sE *SimEnv) Simulate() {
...
+	for sE.dayInfo.CurrTick <= sE.dayInfo.TotalIterations {
 		missingAgentsMap := t.UpdateMissingAgents()
 		for floor := range missingAgentsMap {
 			for _, agentType := range missingAgentsMap[floor] {
 				sE.createNewAgent(a, t, agentType, floor)
 			}
 		}
-
-		a.SimulationIterate(i)
+		sE.dayInfo.CurrTick++
```
jzzheng22, line 76: Why not just put this into the for loop?

jzzheng22 (changes requested):
Thanks Udai, I like the idea - a few changes would here and there and should be fine.

pkg/utils/globalTypes/Day/day.go
```diff
@@ -0,0 +1,21 @@
+package day_type
```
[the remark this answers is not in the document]
JaafarRammal: Yeah and then the import wouldn't need the `.`

pkg/simulation/simulation.go
udai-arneja: Not currently integrated into ABM, and pretty much the same as putting it into a Tower tick, I guess this is in the overall for loop.

cmd/backend/main.go
```diff
@@ -59,17 +60,23 @@ func main() {
 	log.SetOutput(f)
 	log.SetFormatter(&log.JSONFormatter{})
 
+	// Day Information Struct Init
+	dayInfo := DayInformationNew()
+
 	// can have frontend parameters come go straight into simEnv
```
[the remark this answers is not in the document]
udai-arneja: Instantiation - adding some other parameters there. Guess the comment is a quite redundant lol.

cmd/backend/main.go
```diff
@@ -59,17 +60,23 @@ func main() {
...
+	dayInfo.SimulationDays = 10
+	dayInfo.IterationsPerDay = 24 * 60
+	dayInfo.DaysPerReshuffle = 3
+	dayInfo.IterationsPerReshuffle = dayInfo.IterationsPerDay * dayInfo.DaysPerReshuffle
+	dayInfo.TotalIterations = dayInfo.IterationsPerDay * dayInfo.SimulationDays
 
-	iterationsPerDay := 24 * 60
-	iterations := days * iterationsPerDay
-	reshufflePeriod := daysPerReshuffle * iterationsPerDay
+	// agentParameters - struct
```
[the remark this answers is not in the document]
udai-arneja: Placeholder for doing something similar with Agent Default Information, will change to a TODO.

pkg/simulation/simulation.go
jzzheng22: Never mind, got confused.

I think it's a bit weird to not have the tick increment as the last statement in the loop. If we initialise CurrTick to 1 then move the increment to the end of the loop, would that still work?

cmd/backend/main.go
```diff
@@ -59,17 +60,23 @@ func main() {
 	log.SetOutput(f)
 	log.SetFormatter(&log.JSONFormatter{})
 
+	// Day Information Struct Init
+	dayInfo := DayInformationNew()
```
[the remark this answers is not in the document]
udai-arneja: Is it better to have `day.NewDayInfo` rather than `NewDayInfo` by importing slightly differently. I know for some functions its for identification/clarity of source package but in this case is it self-explainority - I'm unsure what the best practice is.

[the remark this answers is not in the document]
udai-arneja: This would affect imports for the rest of packages so holding off for that ATM.

pkg/simulation/simulation.go
udai-arneja: Sounds good👍

cmd/backend/main.go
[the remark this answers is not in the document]
jzzheng22: I'd stick to `day.NewDayInfo` personally... maybe that will change with experience but for now I think better to keep it as `day.NewDayInfo`

jzzheng22:
Why not just call the folder `day` instead of `dayType`? It's in the folder `globalTypes` so it should be fairly obvious that it's a type without you having to say it's a type in the folder name. Would then save you from having to rename the imports

jzzheng22 (changes requested):
I don't see the point in calling the folder `dayType` - if you change it to just `day` then the imports don't need to be renamed. Don't think clarity is an issue here since it's in a directory called globalTypes.

otherwise, everything else is great, thanks

udai-arneja:
I dislike lint haha. Do you know what Im doing wrong, I think its because the package name and folder name are both day? It worked when it was dayType...

udai-arneja:
Thank you lots, changed locally but maybe it doesn't get pushed?

jzzheng22:
> Thank you lots, changed locally but maybe it doesn't get pushed?

Not sure, maybe you weren't adding the right file(s) to the commit?

cmd/backend/main.go
```diff
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/SOMAS2021/SOMAS2021/pkg/simulation"
+	day "github.com/SOMAS2021/SOMAS2021/pkg/utils/globalTypes/day"
```
jzzheng22, line 11:
```suggestion
	"github.com/SOMAS2021/SOMAS2021/pkg/utils/globalTypes/day"
```

jzzheng22 (changes requested):

jzzheng22 (approved):
LGTM (just an FYI I removed the rename of the import of day in main.go)

jzzheng22:
Happy for you to merge from my side, not sure if you want to wait for others to review though. You can use your judgement here :)

JaafarRammal (approved):

Issue #33, opened 2021-12-10
Create a random agent for experimentation


JaafarRammal:
Create a random agent which:
- Does not send messages (?)
- Takes a random amount of food

PR #53, opened 2021-12-14
Adding Random Agent

Rhys-J:
Adding an agent that takes a random amount of food between 0 and 100. Currently does nothing else.

# Summary

Solves #33. Creates a random agent for experimentation

## Additional Information

Random agent that only takes a food amount between 0 and 100. Does not sent messages

## Test Plan

Ran simulation and checked log files to see amount of food taken was random.

### DIFF
--- a/pkg/agents/randomAgent/agent.go
+++ b/pkg/agents/randomAgent/agent.go
@@ -1,0 +1,25 @@
+package agentRandom
+
+import (
+	"math/rand"
+
+	"github.com/SOMAS2021/SOMAS2021/pkg/infra"
+	"github.com/SOMAS2021/SOMAS2021/pkg/utils/abm"
+)
+
+type CustomAgentRandom struct {
+	*infra.Base
+	// new params
+}
+
+func New(baseAgent *infra.Base) (abm.Agent, error) {
+	//create other parameters
+	return &CustomAgentRandom{
+		Base: baseAgent,
+	}, nil
+}
+
+func (a *CustomAgentRandom) Run() {
+	a.Log("Random agent reporting status:", infra.Fields{"floor": a.Floor(), "hp": a.HP()})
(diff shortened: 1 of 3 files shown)


pkg/agents/randomAgent/agent.go
```diff
@@ -0,0 +1,25 @@
+package agentRandom
```
JaafarRammal, line 1: LGTM! Otherwise can we rename just to have consistency?
```suggestion
package randomAgent
```

pkg/simulation/simulation.go
```diff
@@ -1,6 +1,7 @@
 package simulation
 
 import (
+	agentRandom "github.com/SOMAS2021/SOMAS2021/pkg/agents/randomAgent"
```
JaafarRammal, line 4:
```suggestion
	"github.com/SOMAS2021/SOMAS2021/pkg/agents/randomAgent"
```

pkg/simulation/simulation.go
```diff
@@ -89,7 +90,7 @@ func sum(inputList []int) int {
 func (sE *SimEnv) createNewAgent(a *abm.ABM, tower *infra.Tower, i, floor int) {
 	// TODO: clean this looping, make a nice abs map
 	sE.Log("Creating new agent")
-	abs := []AgentNewFunc{agent1.New, agent2.New, team6.New}
+	abs := []AgentNewFunc{agent1.New, agent2.New, team6.New, agentRandom.New}
```
JaafarRammal, line 93:
```suggestion
	abs := []AgentNewFunc{agent1.New, agent2.New, team6.New, randomAgent.New}
```

JaafarRammal (changes requested):
Thanks for the agent! Just a minor rename comment otherwise LGTM

udai-arneja:
Try not to change default parameters in the main.go file when committing, I think its good practice not to because it can get confused with the actual change in the PR. Otherwise LGTM, thanks !

Issue #43, opened 2021-12-11
Move abm to simulation


jzzheng22:
Move required abm code to simulation and remove unnecessary abm code

Issue #20, opened 2021-12-09
Move replacement of dead agents into a separate function


jzzheng22:
`simulationLoop()` in `pkg/infra/simulation/simulation.go`

udai-arneja:
See branch infra-abm-compression

Issue #14, opened 2021-12-09
Change map of agents to map of pointers to agent struct


jzzheng22:
In `pkg/infra/tower/tower.go`

This should make the setter functions a lot cleaner.

udai-arneja:
The Tower Map of agents is a pointer to base agents, so I guess this is referring to the ABM implementation. See branch infra-abm-compression for changes, I think it's addressing this issue? Could be wrong though.

PR coming after a little more cleanup.

jzzheng22:
Before it was a map of structs, not map of pointers to structs - I'll take a look on Thursday

udai-arneja:
Maybe Gowoon fixed it in the rework hehe. Good luck tomo 😄

PR #51, opened 2021-12-14
Day & Tower Input Parameters.

udai-arneja:
Ticks per day now dependent on other variables.

# Summary

This PR is aiming to solve the issue #13 . Day and tower parameters as set from discussion in meeting. 

## Additional Information

From meeting: input params: composition of agents (and therefore total number of agents), number of agents per floor (this will need to be a factor of the total number of agents), time spent per floor. (Courtesy of @jzzheng22 )

## Test Plan

Tested by observing logs of reshuffling etc while changing `ticksPerFloor` in main.go - changes day length as expected.

PR #55, opened 2021-12-15
Infra abm compression

udai-arneja:
# Summary

First of a few big structural changes. Integrated ABM into simulation - with simulation package now across three files. I think the names might need improving. I believe this solves issues: #43 #20 #14 .

## Additional Information

Added all ABM types into globalType declarations.
Added list of custom agents in ABM to map of custom agents (key : uuid) in simulation. - can extend this in the future to also include base agents although I like it as is atm, open for dicussion.

Believe its cleaner now, but am open to moving some functions around etc. I think theres nothing too redundant atm either, but worth checking.

Could someone double check my pointers for custom agent (simulation and `AgentsRun()` in sim-iteration.go ), although it all seems to be working in the test logs.

## Test Plan

Checked for normal behaviour including killing and running.


pkg/simulation/simulation.go
```diff
@@ -1,83 +1,64 @@
...
 
 	totalAgents := sum(sE.AgentCount)
-	t := infra.NewTower(sE.FoodOnPlatform, totalAgents, 1, sE.dayInfo)
-	a.SetWorld(t)
+	t := infra.NewTower(sE.FoodOnPlatform, totalAgents, sE.AgentsPerFloor, sE.dayInfo)
+	sE.SetWorld(t)
 
 	agentIndex := 1
 	for i := 0; i < len(sE.AgentCount); i++ {
 		for j := 0; j < sE.AgentCount[i]; j++ {
```
jzzheng22, line 51: Can I be annoying and ask for you to put agent creation into a nice function too <3

pkg/simulation/sim-iteration.go
```diff
@@ -0,0 +1,57 @@
...
+		sE.AgentsRun()
+		if sE.reportFunc != nil {
+			sE.reportFunc(sE)
+		}
+		sE.dayInfo.CurrTick++
+	}
+}
+
+func (sE *SimEnv) SetReportFunc(fn func(*SimEnv)) {
+	sE.reportFunc = fn
```
jzzheng22, line 34: Is this supposed to be called anywhere?

pkg/simulation/sim-agent.go
```diff
@@ -0,0 +1,45 @@
...
+		for _, agentType := range missingAgentsMap[floor] {
+			sE.createNewAgent(t, agentType, floor)
+		}
+	}
+}
+
+func (sE *SimEnv) AgentsCount(t *infra.Tower) int {
+	sE.mx.RLock()
+	defer sE.mx.RUnlock()
+	return t.TotalAgents()
```
jzzheng22, line 44: Haven't tested, but is this mutex necessary?

pkg/simulation/simulation.go
```diff
@@ -1,83 +1,64 @@
...
+	"github.com/SOMAS2021/SOMAS2021/pkg/utils/globalTypes/agent"
 	"github.com/SOMAS2021/SOMAS2021/pkg/utils/globalTypes/day"
-	"github.com/google/uuid"
+	"github.com/SOMAS2021/SOMAS2021/pkg/utils/globalTypes/world"
 	log "github.com/sirupsen/logrus"
 )
 
 type Fields = log.Fields
 
+type AgentNewFunc func(base *infra.Base) (agent.Agent, error)
```
jzzheng22, line 15: This should move into sim-agent.go

pkg/simulation/sim-agent.go
```diff
@@ -0,0 +1,45 @@
...
+	sE.Log("Creating new agent")
+	abs := []AgentNewFunc{agent1.New, agent2.New, team6.New}
+
+	uuid := uuid.New().String()
+	bagent, err := infra.NewBaseAgent(sE.world, i, sE.AgentHP, floor, uuid)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	custagent, err := abs[i](bagent)
```
jzzheng22, line 23: I'm not sure this is the best way of writing the code - I think it better for the custom agent's constructor to call the base agent constructor, rather than calling the base agent constructor directly here... but I'm happy for that to be fixed in a later PR. Let me know what you decide here and I can create an issue - I'm happy to just fix it myself when I have some free time since it's a not a critical issue.

pkg/simulation/sim-agent.go
```diff
@@ -0,0 +1,45 @@
...
+
+	sE.custAgents[uuid] = custagent
+	tower.AddAgent(bagent)
+}
+
+func (sE *SimEnv) replaceAgents(t *infra.Tower) {
+	missingAgentsMap := t.UpdateMissingAgents()
+	for floor := range missingAgentsMap {
+		for _, agentType := range missingAgentsMap[floor] {
+			sE.createNewAgent(t, agentType, floor)
```
jzzheng22, line 36: Partially a note for myself here - I'd like to see missingAgents removed completely eventually since it's another piece of state that we need to keep track of - better to just check HP directly when we need to. (see #15 ). 

But again, that's an issue for another PR, but mentioning it so we still remember it's an issue.

pkg/simulation/sim-agent.go
```diff
@@ -0,0 +1,45 @@
+package simulation
+
+import (
+	"github.com/SOMAS2021/SOMAS2021/pkg/agents/team1/agent1"
```
jzzheng22, line 4: Can you rename this file to `simAgent.go` please?

pkg/simulation/sim-iteration.go
```diff
@@ -0,0 +1,57 @@
+package simulation
+
+import (
+	"sync"
```
jzzheng22, line 4: Can you rename this file to simIteration.go please?

jzzheng22 (changes requested):
Mostly pretty good, thanks a lot.

Obviously will need to be rebased once #51  is merged, but other than that and a few changes here it looks fine.

Also - can you rename the files so no dashes?

pkg/simulation/sim-iteration.go
udai-arneja: Loool I've got not idea - I've baso copied and pasted it from ABM iterate lol.

pkg/simulation/sim-agent.go
udai-arneja: Will test it and let you know too

pkg/simulation/sim-iteration.go
jzzheng22: In that case, I think delete it - no point in having it unless we know we need it.

We can always add it back later.

pkg/simulation/sim-agent.go
JaafarRammal: > I think it better for the custom agent's constructor to call the base agent constructor

afaik we avoided that because then the custom constructor can manipulate / break the base agent. Since base agent has the following properties used for state control:

```
id        string
hp        int
floor     int
inbox     *list.List
tower     *Tower
```

Exposing the constructor means exposing things like tower since we need to pass it to custom for a successful constructor.

udai-arneja: Yeah agreed with Jaafar here, happy to hear further thoughts though Jason @jzzheng22

pkg/simulation/sim-agent.go
udai-arneja: Agreed, I think something to come back to after the HPdecay has been implemented properly?

pkg/simulation/sim-agent.go
jzzheng22: Low priority anyway - happy to keep this in for now but I'll probably think about it over the break

udai-arneja:
Could you check if a mutex is needed around line 43 in simAgent? also I think the other mutex when returning t.TowerAgents() is needed.

pkg/simulation/simAgent.go
```diff
@@ -0,0 +1,60 @@
...
+	bagent, err := infra.NewBaseAgent(sE.world, i, sE.AgentHP, floor, uuid)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	custagent, err := abs[i](bagent)
+	if err != nil {
+		log.Fatal(err)
+	}
+	// TODO: mutex here?
```
jzzheng22, line 42: I don't think we'll need a mutex here - it's not called inside a goroutine so it's not running concurrently.
```suggestion
```

pkg/simulation/simAgent.go
```diff
@@ -0,0 +1,60 @@
...
+		for _, agentType := range missingAgentsMap[floor] {
+			sE.createNewAgent(t, agentType, floor)
+		}
+	}
+}
+
+func (sE *SimEnv) AgentsCount(t *infra.Tower) int {
+	sE.mx.RLock()
+	defer sE.mx.RUnlock()
+	return t.TotalAgents()
```
jzzheng22, line 59: Wouldn't you have to lock the tower mutex instead of this one?

jzzheng22 (changes requested):

JaafarRammal:
@udai-arneja 

First mutex in simulation around line 43 is needed. Say we decided to multithread the function call, or there is two concurrent calls to the function with the same sE, or an access to the resource while we are writing to it, then we'd need to lock the resource.

I am not sure about the 2nd one but I think we do need the `RLock()` on it

pkg/simulation/simAgent.go
JaafarRammal: Correct

pkg/simulation/simAgent.go
JaafarRammal: Say we decided to multithread the function call, or there is two concurrent calls to the function with the same sE, or an access to the resource while we are writing to it, then we'd need to lock the resource.

I am not sure about the 2nd one but I think we do need the RLock() on it

JaafarRammal:
@jzzheng22 is #55 resolved by this PR?

JaafarRammal (commented):
LGTM, just mutex change as mentioned by Jason

pkg/simulation/simAgent.go
jzzheng22: > Say we decided to multithread the function call, or there is two concurrent calls to the function with the same sE, or an access to the resource while we are writing to it, then we'd need to lock the resource.

When would that ever happen though? Are we planning to make this part run concurrently??

jzzheng22 (approved):
I assume it's fine... we'll need to start enforcing unit tests in PRs soon though to make things easier to review.

PR #57, opened 2021-12-17
Order of Loop - agent run first then tower tick().

udai-arneja:
Co-Authored-By: Priya Chhaya <77248784+priyachh@users.noreply.github.com>

# Summary

First the agent Run then the tower ticks. This makes more sense so the reshuffling doesn't happen before the agents have their last run.

## Additional Information

Also changed around logs a little so its gives better info.

## Test Plan

From the logs, they are more coherent. Day and tower resets occur at the right times.


mdubied:
1) Current Simulation Tick should appear in the first position
2) Printing a Current Simulation Day would be nice to have

mdubied:
to improve readibility:

- [ ] add a blank line or just a line (----------------------) at the end of every day

- [ ] only display "msg":"Reporting platform status end of day" at the end of the day and not at every tick

pkg/infra/tower.go
```diff
@@ -46,9 +50,9 @@ func NewTower(maxPlatFood float64, agentCount,
 
 func (t *Tower) Tick() {
 	//logs
-	t.Log("", Fields{"Curr Tower Tick": t.dayInfo.CurrTick})
-	t.Log("Reporting platform status", Fields{"food_left": t.currPlatFood, "floor": t.currPlatFloor})
-
+	t.mx.RLock()
+	t.TowerStateLog(" end of day")
+	t.mx.RUnlock()
```
jzzheng22, line 55: I'm not sure we need this mutex - I don't think Tick() is ever called by multiple threads?

pkg/simulation/simAgent.go
```diff
@@ -50,6 +49,7 @@ func (sE *SimEnv) replaceAgents(t *infra.Tower) {
 			sE.createNewAgent(t, agentType, floor)
 		}
 	}
+
```
jzzheng22, line 52:
```suggestion
```

jzzheng22 (changes requested):
Couple of questions but otherwise it's good, thanks

pkg/infra/tower.go
udai-arneja: Truuuuee

pkg/infra/tower.go
```diff
@@ -123,6 +127,7 @@ func (t *Tower) SendMessage(direction int, senderFloor int, msg messages.Message
 func (t *Tower) ResetTower() {
 	t.currPlatFood = t.maxPlatFood
 	t.currPlatFloor = 1
+	t.Log("Tower Reset", Fields{"currFood": t.currPlatFood, "maxFood": t.maxPlatFood, "currFloor": t.currPlatFloor})
```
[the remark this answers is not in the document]
udai-arneja: True guess we dont need to print it but its worth having a log saying reset imo?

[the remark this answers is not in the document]
jzzheng22: Yup, happy to print the reset, but don't need to pollute the logs with the rest of the info

pkg/infra/tower.go
```diff
@@ -64,6 +66,7 @@ func (t *Tower) Tick() {
 	if t.dayInfo.CurrTick%t.dayInfo.TicksPerDay == 0 {
 		t.hpDecay() // decreases HP and kills if < 0
 		t.ResetTower()
+		t.Log("-----------------E-N-D----O-F----D-A-Y-----------------", Fields{})
```
jzzheng22, line 69:
```suggestion
		t.Log("-----------------END----OF----DAY-----------------", Fields{})
```

pkg/infra/tower.go
```diff
@@ -46,9 +50,7 @@ func NewTower(maxPlatFood float64, agentCount,
 
 func (t *Tower) Tick() {
 	//logs
```
jzzheng22, line 52:
```suggestion
```

jzzheng22: The fact that you're calling the function `TowerStateLog()` should hopefully make it pretty clear that it is a log

jzzheng22 (changes requested):

mdubied (changes requested):
See my comments above

udai-arneja:
@mdubied I think having it at the start and end of every tick is better for individual log reading?

mdubied (dismissed):

mdubied:
@udai-arneja the line you mean? I would keep it at the end of the day for now. Or single line for End of the tick, double line for end of the day

udai-arneja:
@mdubied so keep it as it is. there is a. platform status before and after each tick and then a dashed line at the end of each day.

jzzheng22 (approved):

jzzheng22 (approved):

PR #62, opened 2021-12-18
Health

mdubied:
# Summary

This PR aims to implement two more advances health functions: hpDecay and updateHP. The former takes care of lowering the hp value at the end of the day. The latter increases the hp when agents eat.

## Additional Information

Most details about the implementation are given in slack (health channel). The important changes are located in baseagent.go, tower.go, and health.go. Documentation will be available soon.

Ask me if you have any questions, or if something is unclear.

## Test Plan

Tested several starting setup and TakeFood value for agent1. Comparison with the analytical curves.


JaafarRammal:
https://github.com/SOMAS2021/SOMAS2021/actions/runs/1596979712
Pipeline is passing, GitHub is just not displaying things correctly not sure why

pkg/infra/baseagent.go
```diff
@@ -93,8 +94,34 @@ func (a *Base) setHP(newHP int) {
 	a.hp = newHP
 }
 
-func (a *Base) updateHP(foodTaken float64) {
-	a.hp = int(math.Min(100, float64(a.hp)+foodTaken))
+func (a *Base) updateHP(foodTaken float64) { // first order system step answer. 63% of levelWidth obtained for tau food
```
jzzheng22, line 97: I don't know what this comment means. In any case, it should go directly below the function prototype anyway.

pkg/infra/baseagent.go
```diff
@@ -93,8 +94,34 @@ func (a *Base) setHP(newHP int) {
...
+		a.hp = a.tower.healthInfo.WeakLevel + int(a.tower.healthInfo.WidthWeak*(1-math.Pow(math.E, -(foodTaken+shift)/tau)))
+
+	// Critical Level - TODO: discuss its implementation
+	case a.hp >= 0:
+		if foodTaken >= a.tower.healthInfo.FoodReqCToW {
+			a.hp = int(a.tower.healthInfo.WeakLevel) - 1
+		} else {
+			a.hp = 1
+		}
+	}
```
jzzheng22, line 124:
```suggestion
	// tau is the time constant of the first order system
	tau := 0.0
	base := 0
	width := 0.0

	switch {
	case a.hp >= a.tower.healthInfo.StrongLevel:
		tau = a.tower.healthInfo.FoodReqStrong 
		base = a.tower.healthInfo.StrongLevel
		width = a.tower.healthInfo.WidthStrong

	case a.hp >= a.tower.healthInfo.HealthyLevel:
		tau = a.tower.healthInfo.FoodReqHealthy
		base = a.tower.healthInfo.HealthyLevel
		width = a.tower.healthInfo.WidthHealthy

	case a.hp >= a.tower.healthInfo.WeakLevel:
		tau = a.tower.healthInfo.FoodReqWeak
		base = a.tower.healthInfo.WeakLevel
		width = a.tower.healthInfo.WidthWeak

	// Critical Level - TODO: discuss its implementation
	case a.hp >= 0:
		if foodTaken >= a.tower.healthInfo.FoodReqCToW {
			a.hp = int(a.tower.healthInfo.WeakLevel) - 1
		} else {
			a.hp = 1
		}
		return
	}

	a.hp = int(float64(base) + width - math.Pow(math.E, -foodTaken/tau) * float64(1 - a.hp + base))
```

pkg/infra/baseagent.go
```diff
@@ -14,15 +14,16 @@ import (
...
+	id                string
+	hp                int
+	floor             int
+	agentType         int
+	inbox             *list.List
+	tower             *Tower
+	mx                sync.RWMutex
+	logger            log.Entry
+	hasEaten          bool
+	numberInStateDays int //initialized to 0 by default
```
jzzheng22, line 26:
```suggestion
	daysInState int
```

jzzheng22: Will need to change other instances of this too.

pkg/utils/globalTypes/health/health.go
```diff
@@ -0,0 +1,33 @@
+package health
+
+type HealthInfo struct {
+	StrongLevel    int // defined as the lowest HP value for a given HP level
```
jzzheng22, line 4:
```suggestion
	// Base HP for each health level
	StrongLevel    int 
```

pkg/utils/globalTypes/health/health.go
```diff
@@ -0,0 +1,33 @@
...
+	HealthyLevel   int
+	WeakLevel      int
+	WidthStrong    float64 // width of every level
```
jzzheng22, line 7:
```suggestion
	// Width of every health level
	WidthStrong    float64 
```

pkg/utils/globalTypes/health/health.go
```diff
@@ -0,0 +1,33 @@
...
+	WidthHealthy   float64
+	WidthWeak      float64
+	WidthCritical  float64
+	FoodReqStrong  float64 // food requirement to stay at a given level
```
jzzheng22, line 11:
```suggestion
	// Food intake required to stay at a given level
	FoodReqStrong  float64 
```

pkg/utils/globalTypes/health/health.go
```diff
@@ -0,0 +1,33 @@
...
+	FoodReqHealthy float64
+	FoodReqWeak    float64
+	FoodReqCToW    float64
+	MaxDayCritical int // maximum number of day at critical level
```
jzzheng22, line 15:
```suggestion
	// Number of days an agent can stay critical before dying
	MaxDayCritical int
```

pkg/infra/baseagent.go
jzzheng22: I worked through the formula you've given in the code for the hp calculation at the end... hopefully it's right. I've tried to simplify it so there's not as much computation required.

jzzheng22 (changes requested):
This review is from a code review perspective, not an implementation perspective - I'm still not 100% satisfied with the decision making here, but I'll be happier once I understand the reasoning behind it.

cmd/backend/main.go
```diff
@@ -62,19 +63,32 @@ func main() {
 
 	// can have frontend parameters come go straight into simEnv
 	foodOnPlatform := 100.0
-	numOfAgents := []int{2, 2, 2, 2, 2}
-	agentHP := 100
+	numOfAgents := []int{2, 0, 0, 0, 0} //agent1, agent2, team3, team6, randomAgent
```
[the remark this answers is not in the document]
mdubied: Thiis was for testing purposes only, we can go back to 2,2,2,2,2

cmd/backend/main.go
```diff
@@ -62,19 +63,32 @@ func main() {
...
+	agentHP := 55
```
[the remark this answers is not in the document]
mdubied: Again for testing purposes. I will accept the change. But be aware that agentHP is not maxHP! agentHP is the initialization of the agents HP, at the beginning of the simulation

cmd/backend/main.go
```diff
@@ -62,19 +63,32 @@ func main() {
...
+	//numberCriticalDays := 0
 	agentsPerFloor := 1 //more than one not currently supported
 	numberOfFloors := simulation.Sum(numOfAgents) / agentsPerFloor
-	ticksPerFloor := 10
+	ticksPerFloor := 1
```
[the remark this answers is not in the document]
mdubied: Testing purposes agai. We can switch back to 10.

pkg/infra/tower.go
```diff
@@ -98,11 +102,62 @@ func (t *Tower) reshuffle(numOfFloors int) {
 }
 
 func (t *Tower) hpDecay() {
-	// TODO: can add a parameter
 	for _, agent := range t.agents {
-		newHP := agent.hp - 20
+		var newHP int
```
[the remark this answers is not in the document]
mdubied: I think that int are always initialize with 0 in go, but I will commit this change, this is probably more readable

cmd/backend/main.go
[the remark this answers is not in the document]
udai-arneja: Should undo changes like this because it'll create unnecessary merge conflicts down the road & isn't relevant to the commit @mdubied

cmd/backend/main.go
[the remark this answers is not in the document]
mdubied: agentHP is now set back to 100. I kept the name agentHP, as it is not necessarily the same as maxHP

cmd/backend/main.go
```diff
@@ -62,19 +63,32 @@ func main() {
 
 	// can have frontend parameters come go straight into simEnv
 	foodOnPlatform := 100.0
-	numOfAgents := []int{2, 2, 2, 2, 2}
-	agentHP := 100
+	numOfAgents := []int{2, 0, 0, 0, 0} //agent1, agent2, team3, team6, randomAgent
+	agentHP := 55
+	//numberCriticalDays := 0
```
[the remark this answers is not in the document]
mdubied: commit this change

cmd/backend/main.go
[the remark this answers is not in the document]
mdubied: Done: ticksPerFloor changed to 10.

pkg/infra/tower.go
```diff
@@ -98,11 +102,62 @@ func (t *Tower) reshuffle(numOfFloors int) {
...
+		switch {
+		// Strong Level
```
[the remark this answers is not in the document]
mdubied: Isn't it easier to read the code with the comment? It make the file a bit longer I have to admit.

pkg/infra/tower.go
```diff
@@ -98,11 +102,62 @@ func (t *Tower) reshuffle(numOfFloors int) {
...
+		case agent.hp >= t.healthInfo.StrongLevel:
+			if agent.hp >= t.healthInfo.StrongLevel+int(0.63*t.healthInfo.WidthStrong) {
+				newHP = int(math.Max(float64(t.healthInfo.StrongLevel), float64(agent.hp)-0.63*t.healthInfo.WidthStrong))
+				agent.numberInStateDays++
+			} else {
+				newHP = int(math.Max(float64(t.healthInfo.HealthyLevel), float64(agent.hp)-0.63*t.healthInfo.WidthStrong))
+				agent.numberInStateDays = 0
+			}
+
+		// Healthy Level
```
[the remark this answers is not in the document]
mdubied: Isn't it easier to read the code with the comment? It make the file a bit longer I have to admit.

pkg/infra/tower.go
```diff
@@ -98,11 +102,62 @@ func (t *Tower) reshuffle(numOfFloors int) {
...
+				agent.numberInStateDays = 0
+			case agent.hp >= t.healthInfo.HealthyLevel+int(0.63*t.healthInfo.WidthHealthy):
+				newHP = int(math.Max(float64(t.healthInfo.HealthyLevel), float64(agent.hp)-0.63*t.healthInfo.WidthHealthy))
+				agent.numberInStateDays++
+			default:
+				newHP = int(math.Max(float64(t.healthInfo.WeakLevel), float64(agent.hp)-0.63*t.healthInfo.WidthHealthy))
+				agent.numberInStateDays = 0
+			}
+
+		// Weak Level
```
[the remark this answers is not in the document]
mdubied: Isn't it easier to read the code with the comment? It make the file a bit longer I have to admit.

pkg/utils/globalTypes/health/health.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+func NewHealthInfo(StrongLevel, HealthyLevel, WeakLevel int, FoodReqStrong, FoodReqHealthy, FoodReqWeak, FoodReqCToW float64, MaxDayCritical int) *HealthInfo {
+	return &HealthInfo{
+		StrongLevel:    StrongLevel,
+		HealthyLevel:   HealthyLevel,
+		WeakLevel:      WeakLevel,
+		WidthStrong:    float64(100 - StrongLevel),
```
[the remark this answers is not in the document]
mdubied: I don't think so. If I understand correctly, `agentHP` is the initial HP value of every agent. We can choose to set this initial value to the maximal HP (100), but we can also choose to initialize it with any other value (for example every agent critical at the beginning of the simulation). This has therefore to be an independent variable in my opinion

pkg/utils/globalTypes/health/health.go
```diff
@@ -0,0 +1,33 @@
...
+		WidthHealthy:   float64(StrongLevel - HealthyLevel),
+		WidthWeak:      float64(HealthyLevel - WeakLevel),
+		WidthCritical:  float64(WeakLevel),
```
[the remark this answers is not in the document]
mdubied: The function `updateHP` need to use floats for its computations. I thought it was better to let these 4 variables as float as we don't need to cast them afterwards.

If we use the step response 1st order systems (or any other exponential curve) as a way to model `updateHP`, we will need to use floats. The problem comes rather from the fact that `hp` is an int. This is maybe something that we have to discuss with the infra team @JaafarRammal @udai-arneja (only tagging the members that review this PR)

pkg/infra/baseagent.go
woonmoon: +1 @jzzheng22 's implementation however could you make some clear documentation about the formula ahead of every case statement
`// HP = Level + Width * (1 - e^tau)`
Something like this on top maybe? And better documentation on what `Level`, `Width` and `tau` denote exactly.

pkg/infra/tower.go
```diff
@@ -98,11 +102,62 @@ func (t *Tower) reshuffle(numOfFloors int) {
...
+		switch {
+		// Strong Level
+		case agent.hp >= t.healthInfo.StrongLevel:
+			if agent.hp >= t.healthInfo.StrongLevel+int(0.63*t.healthInfo.WidthStrong) {
+				newHP = int(math.Max(float64(t.healthInfo.StrongLevel), float64(agent.hp)-0.63*t.healthInfo.WidthStrong))
+				agent.numberInStateDays++
+			} else {
+				newHP = int(math.Max(float64(t.healthInfo.HealthyLevel), float64(agent.hp)-0.63*t.healthInfo.WidthStrong))
+				agent.numberInStateDays = 0
+			}
```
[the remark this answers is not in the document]
woonmoon: I think hardcoding numbers is fine when we do not intend to parameterize it.

pkg/infra/baseagent.go
mdubied: Good suggestion. I commit it but will slightly change the function line 129: you forgot the "shift". This shift might seem "magical" but it's not. I'll document its derivation and post it on slack. It is linked to the geogebra file + explanation I gave in slack.

pkg/infra/tower.go
```diff
@@ -98,11 +102,62 @@ func (t *Tower) reshuffle(numOfFloors int) {
 }
 
 func (t *Tower) hpDecay() {
-	// TODO: can add a parameter
 	for _, agent := range t.agents {
-		newHP := agent.hp - 20
+		newHP := 0
+		switch {
+		// Strong Level
```
woonmoon, line 108: I think hardcoding 0.63 is fine as we don't intend to parameterize further down the line. However, similarly to my comment on `basegent.go` if there could be better documentation for the formula/reasoning for `newHP` 
I know we won't be maintaining this codebase for very long so it's less for maintenance sake and more about clarity, IMO if at a later point we want to change this formula, we shouldn't have to wonder why these numbers were picked in the first place.

pkg/utils/globalTypes/health/health.go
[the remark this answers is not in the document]
woonmoon: Nothing decisive to add but I was under the impression we would never vary initial HP of an agent in which case I agree with @jzzheng22

woonmoon (commented):
Almost ready for merge? IMO might be worthwhile to add documentation where numbers are hardcoded/the formula case-switch statements.

pkg/utils/globalTypes/health/health.go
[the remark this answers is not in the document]
mdubied: I think it could be interesting to try different initial conditions for the agent health, and would therefore keep it as a parameter. A society that begins with only "strong" agents has more time to discuss treaties etc in comparison to a society beginning with agent that are already in "critical" state from day 1
Also, this parameter is quite convenient to quickly test different scenarios when implementing a new feature: in my case, I could directly start the simulation at the critical level to check if my specific changes were correctly implemented for example

pkg/infra/tower.go
mdubied: Documentation is definitely needed you are right. I will begin to write the documentation when I am done with all requests. 
@woonmoon Do you think I should add more comments directly in the text in addition of the documentation? Or is the documentation sufficient to explain modeling+hard coded number

pkg/infra/baseagent.go
mdubied: I corrected the comment placement. More information will be given in the documentation

pkg/infra/tower.go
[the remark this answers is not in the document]
jzzheng22: The code is comparing hp with `StrongLevel` though - I think that's clear enough

pkg/infra/tower.go
jzzheng22: Hmm okay, happy to keep the magic numbers then. But the meanings need to be documented in the code (and probably elsewhere too).

pkg/infra/tower.go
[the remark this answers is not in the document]
jzzheng22: See above

pkg/infra/tower.go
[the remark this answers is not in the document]
jzzheng22: See above

pkg/utils/globalTypes/health/health.go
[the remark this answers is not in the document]
jzzheng22: I'm happy to keep `initialHP` and `maxHP` as separate variables, but in either case, 100 should not be there: it should be either `initialHP` (i.e. `agentHP`) or `maxHP`. 100 is a magic number that just appears

pkg/utils/globalTypes/health/health.go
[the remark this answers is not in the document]
jzzheng22: It's weird to mix between floats and ints is all - I'd rather just cast where necessary

pkg/infra/tower.go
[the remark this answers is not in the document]
jzzheng22: True, in this case it's probably fine - but needs to be documented.

pkg/infra/baseagent.go
jzzheng22: I'm not sure I agree. I didn't forget the shift: I substituted the formula for shift and simplified the final calculation.

pkg/infra/tower.go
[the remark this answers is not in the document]
mdubied: Documentation will follow shortly

cmd/backend/main.go
```diff
@@ -62,19 +63,31 @@ func main() {
...
 
+	// define heath parameters
+	strongLevel := 55
+	healthyLevel := 25
+	weakLevel := 5 // larger than maxCriticalDay
+	maxDayCritical := 3
+	foodReqStrong := 20.0
+	foodReqHealthy := 15.0
+	foodReqWeak := 10.0
+	foodReqCToW := 2.0
```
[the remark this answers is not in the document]
udai-arneja: Also what are opinions on putting these straight into the NewHealth struct?

udai-arneja:
Why does simulation have a pointer to healthInfo? Doesn't seem necessary unless I'm missing something - its just used in the tower so can it not be instantiated there? @mdubied

pkg/utils/globalTypes/health/health.go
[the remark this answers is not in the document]
mdubied: I think that 100 as a value of `maxHP` will actually not change. I thought we agreed on that earlier but I might be wrong. I would say we can now set 100 as a hard-coded maximal HP value, I don't really see for which purpose we would play with this value in the future

pkg/simulation/simulation.go
```diff
@@ -20,16 +21,18 @@ type SimEnv struct {
 	AgentsPerFloor int
 	logger         log.Entry
 	dayInfo        *day.DayInfo
+	healthInfo     *health.HealthInfo
 	world          world.World
 	custAgents     map[string]agent.Agent
 }
 
-func NewSimEnv(foodOnPlat float64, agentCount []int, agentHP, agentsPerFloor int, dayInfo *day.DayInfo) *SimEnv {
+func NewSimEnv(foodOnPlat float64, agentCount []int, agentHP, agentsPerFloor int, dayInfo *day.DayInfo, healthInfo *health.HealthInfo) *SimEnv {
```
udai-arneja, line 29: Refer to comment on necessity on having this simulation, rather than instantiation in tower directly.
If for input from front end to change these params for different 'difficulties' of simulation, I would argue its better to have three (arbitrary) different levels that the front end can choose between and these are passed into the tower.

udai-arneja: Comment: Why does simulation have a pointer to healthInfo? Doesn't seem necessary unless I'm missing something - its just used in the tower so can it not be instantiated there?

cmd/backend/main.go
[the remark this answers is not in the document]
mdubied: Indeed, it actually changes depending on where you are in the bracket. These values are only valid if your hp is at the lower border of a given HP level. It can actually be misleading, maybe I should change the name. It is basically the parameter "tau" of the equation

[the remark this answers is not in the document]
mdubied: We could put them strain into the NewHealth struct yes. If the others agree I can change that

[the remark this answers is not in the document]
udai-arneja: Ah understood, yeah I think straight in is better. They don't seem like they'll be variable once finalised. Opinions welcome @JaafarRammal @woonmoon @jzzheng22

pkg/utils/globalTypes/health/health.go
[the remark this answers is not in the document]
hussain2603: I agree with @jzzheng22. instead of writing '100' here, we should probably code it as 'maxHP'. It kind of seems that 100 is a random magic number that appeared out of nowhere, but writing 'maxHP' would make it easier to see how widthStrong is being calculated (i.e. maxHP - StrongLevel).

hussain2603 (commented):
just some comments in some places. looks good so far.

pkg/utils/globalTypes/health/health.go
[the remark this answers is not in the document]
mdubied: Done, MaxHP is now a variable of the struct HealthInfo. Its value is set at the beginning of the simulation in main.go.

pkg/utils/globalTypes/health/health.go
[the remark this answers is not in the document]
mdubied: Done: the different LevelWidth are now int. They are casted to float when needed

cmd/backend/main.go
```diff
@@ -79,12 +80,12 @@ func main() {
 	healthyLevel := 25
 	weakLevel := 5 // larger than maxCriticalDay
 	maxDayCritical := 3
-	foodReqStrong := 20.0
-	foodReqHealthy := 15.0
-	foodReqWeak := 10.0
+	tauStrong := 20.0
+	tauHealthy := 15.0
+	tauWeak := 10.0
```
hussain2603, line 85: I think I preferred the old names for tauStrong, tauHealthy, tauWeak haha. (this is minor and maybe can just make some comments to explain those variables).

hussain2603 (commented):
Minor comments for the variable naming. The rest of the changes look good to me.

pkg/infra/baseagent.go
mdubied: My bad, I didn't see the substitution. I think there is still a small mistake. Instead of:
`a.hp = int(float64(base) + width - math.Pow(math.E, -foodTaken/tau) * float64(1 - a.hp + base))` ,
it should be
`a.hp = int(float64(base) + width - math.Pow(math.E, -foodTaken/tau) * float64(width - a.hp + base))`.

I corrected this mistake and use your notation. I will also add the whole derivation in the wiki

cmd/backend/main.go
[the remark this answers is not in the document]
mdubied: I changed the them for `tauStrong`, `tauHealthy` and `tauWeak`

pkg/simulation/simulation.go
mdubied: I just followed what was done before for `DayInfo`, as the two structs are quite similar in their use. But I must admit that I am not sure if this pointer is needed for `NewSimEnv`. I think it is actually:
->We create a struct `HealthInfo` in `main.go`, which is then given to `simEnv`. 
->when calling `simEnv.Simulate()`, one of the first step is to create the tower with `t := infra.NewTower(sE.FoodOnPlatform, totalAgents, sE.AgentsPerFloor, sE.dayInfo, sE.healthInfo)` (in `simulation.go)`.
-> we use the struct `HealthInfo` "attached" to the tower in `tower.go`, when calling `hpDecay` (which take the following argument: `func (t *Tower) hpDecay())`, so a point to tower, which in turn contains a `HealthInfo` struct

mdubied: I might be wrong as I did not work on this part of the code. I based my implementation on `DayInfo` regarding its link to tower/simulation as I said.

cmd/backend/main.go
mdubied: hahahh we have some diverging opinion on that. On a theoretical perspective, `tau` makes more sense as this is a characteristic of the curve/function we model. However, on a practical level, `tau` is defined as "the amount of food needed to stay in a given HP level _if you wake up on the lower border of this level_". From this last "if" condition, it is possible for an agent to _eat less than_ `tau` food and still wake up in the same level the following day (if he wakes up with _more_ than just the level's lower border HP value. Therefore, I think `foodReq` can be misleading

pkg/simulation/simulation.go
JaafarRammal: I think this implementation works for parametrisation

JaafarRammal (approved):
LGTM, parametrisable too

hussain2603 (dismissed):

cmd/backend/main.go
hussain2603: yeah makes sense thanks!

cmd/backend/main.go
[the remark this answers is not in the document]
JaafarRammal: @udai-arneja good point. I think we can go with this for now, and if the function needs to be changed for health / experimented with we can add even more parametrisation

pkg/infra/tower.go
```diff
@@ -98,11 +102,59 @@ func (t *Tower) reshuffle(numOfFloors int) {
...
+				agent.daysInState = 0
+			}
+
+		case agent.hp >= t.healthInfo.HealthyLevel:
+			switch {
+			case agent.hp >= t.healthInfo.HealthyLevel+int(0.95*float64(t.healthInfo.WidthHealthy)):
+				newHP = t.healthInfo.StrongLevel
+				agent.daysInState = 0
+			case agent.hp >= t.healthInfo.HealthyLevel+int(0.63*float64(t.healthInfo.WidthHealthy)):
+				newHP = int(math.Max(float64(t.healthInfo.HealthyLevel), float64(agent.hp)-0.63*float64(t.healthInfo.WidthHealthy)))
```
Rhys-J, line 123: Will the new HP not always be the right side value if you're in this case? Please correct me if I'm misunderstanding the code.

pkg/infra/baseagent.go
```diff
@@ -93,8 +95,39 @@ func (a *Base) setHP(newHP int) {
...
+	case a.hp >= 0:
+		if foodTaken >= a.tower.healthInfo.FoodReqCToW {
+			a.hp = int(a.tower.healthInfo.WeakLevel) - 1
+		} else {
+			a.hp = 1
+		}
+		return
+	}
+
+	a.hp = base + width - int(math.Pow(math.E, -foodTaken/tau)*float64(width-a.hp+base))
```
Rhys-J, line 130: If you are midway through a level and then eat the value of tau, you will end the day on less hp than you started. I don't think this is the correct way to implement it. The shifting of the curve as described in the docs means that moving up to a new level is too difficult in my opinion. I give two example below to show why I think this is an issue.

Example 1:
You start at 45 hp
You eat 15 hp (tau in healthy), taking you to 51 hp by the curve.
In hpDecay, you lose 19 hp (0.63*30) taking you to 32 hp at the end of the day.

Example 2:
You start at 45 hp
You eat 20 hp (the value of tau in strong), taking you to 52 hp by the curve
In hp decay, you lose 19 hp taking you to 33 hp at the end of the day.
This means that maintaining hp in the upper section of the healthy range is more difficult than in the strong stage, which I don't think should be the case. 

I think this needs further discussion unless others disagree with this being an issue or I have completely misinterpreted the code, if this is the case then please correct me. Otherwise, I think this needs a rework.

pkg/infra/tower.go
SaraFFdez: I think this is true and could be changed to
```suggestion
				newHP = int(float64(agent.hp)-0.63*float64(t.healthInfo.WidthHealthy))
```

SaraFFdez (dismissed):
Overall looks good but I think the issue Rhys has spotted needs to be addressed before merging

pkg/infra/tower.go
mdubied: @Rhys-J @SaraFFdez This was thought as a safeguard if for some reasons we change the parameters defining the levels' limits: for example, if we choose the weak level to be very narrow, substracting `0.63*float64(t.healthInfo.WidthHealthy))` from an agent being in the healthy level could lead this agent to wake up directly in the critical level. This would be the case if `WidthWeak` < `0.63*float64(t.healthInfo.WidthHealthy))` (with an agent starting on the lower border of the healthy level for example).
However, if we leave the parameters defining the levels as they are, this should never be an issue and we can simplify the code as you suggested. 
Let me know what you think about this, and I'll change the code accordingly if needed

pkg/infra/baseagent.go
mdubied: This is indeed a very good point, thank you for the examples; they help to understand your thoughts. 
The second example is for me the most relevant. I agree that maintaining your hp in the healthy level (independently to where you are exactly in this level) should not be more difficult than doing so in the strong level. I see two possible ways to fix this:
1) We shift the curves vertically instead of horizontally. In this case the diminishing return is only dependent on the amount of food (and your hp level as it parameterizes the curve), and _not_ on your exact hp value. 
2) Try to find a parametrisation of the hp levels that does not lead to such critical scenarios.

I think 1) makes quite a lot of sense to me. The original idea of shifting the curve horizontally was intended to capture the diminishing return as a function of your hp, but it makes actually more sense to have diminishing returns depending on the amount of food you eat. 1) would capture this.

mdubied: If we decide to implement 1), there is one additional thing we would have to discuss. Right now, the curves saturate at the upper border of each level. this was done so to prevent agents to jump from, for example, weak to strong in one day. Indeed, we want to model the fact that it needs time to "recover" and to feel better. 
If we shift the curves vertically, we would have curves that saturate higher than the level upper boundary. From this, two questions come into my mind:
a) what is the threshold to wake up in the upper level the following day? Is it still as soon as you reach the border of the level ?
b) how do we compute hpDecay if we allow curve to cross the level borders? Do we have something like `max{upperLevel, curentHP-0.63*levelWidth}`, where `levelWidth` is defined as the width of the level the agent started in?
c) What should we do with the amount of food taken by the agent after the curve cross the upper border? Does this food still contribute to increase the agent's hp or do we say something like "you reached the next hp level, you will wake up in this level but your extra food brings you nothing?

pkg/infra/tower.go
mdubied: Nevermind, the safeguard only apply in the last case where
```
default:
				newHP = int(math.Max(float64(t.healthInfo.WeakLevel), float64(agent.hp)-0.63*float64(t.healthInfo.WidthHealthy)))
```
You are completely right about the line you mentioned. This is probably also the case in the other levels. I will correct the code tomorrow.

pkg/infra/baseagent.go
Rhys-J: I think shifting the curves vertically is the right way to do it in this implementation. Changing the max limit of hp you can get to from base+width to a.hp+width would set the upper limit to a more realistic threshold. Then the exponential will be multiplied by the width only. 

```suggestion
	a.hp = a.hp + width - int(math.Pow(math.E, -foodTaken/tau)*float64(width))
```

Maybe this would change the curve to a vertical shift. Not sure what changes need to happen in hpDecay for this to be acceptable though.

I think the limit for hp at the end of the day should not be set by the level thresholds, but by the starting hp plus some maximum gain (width)

mdubied: I just came back home, so I'll have a look at this in details tonight/tomorrow. Your suggestion seems good. We just need to clarify what what when an agent change his level.

mdubied: @Rhys-J I commit your suggestion. It makes sense to me. For updateHP, I also added a safeguard so that the HP can not grow bigger than maxHP.

mdubied: Concerning hpDecay, I left it as it is. It is almost good, but we still have a small problem. Right now, the "cost of living" is always `0.63*currentWidth` (with safeguard in the code to prevent falling too low-normaly not needed with good choice the levelWidth), except if `a.hp>=0.95*currentWidth`. In this last case, the cost of living is 0 and you go up one level. This seems a bit strange with the current vertical shift implementation, as _you could wake up at_ `a.hp=currentLevel+0.95*currentWidth`, _eat nothing, and wake up in the upper level the following day_.

We could avoid this by saying that the cost of living is _really always_ `0.63*currentWidth`, and don't treat this scenario where `a.hp>=0.95*currentWidth` separately. It seems appealing, but this would lead to the fact that it is would take several days for an agent to "level up" starting from the lower border of a level. Let me give you an example:

> You start at the lower border of healthy. **Day 1** you eat and get +0.83*widthHealthy hp. **Day 2** you wake up with `healthyLevel+0.83*healthyWidth-0.63*healthy Width=healthyLevel+0.20*width Healthy`. You eat and get +0.83*widthHealthy hp. **Day 3** You wake up at  `healthyLevel+0.2*healthyWidth+0.83*healthyWidth-0.63*strong Width<=healthyLevel+0.40*width Healthy`(as you substract **strong**Width) Etc etc. Also if you eat 100% of the currentWidth, it takes you a bit of time to level up. >

Therefore, I would probably keep it like this for now, but I am not completely happy with it

mdubied: **Paramatrization of tau**: I also changed the values of tau so that it makes more sense. When running simulations, I saw that with our former parametrisation, eating the same quantity of food in Strong level could actually bring you more hp than in the weak level. This is because the exponential decay term it not big enough and the Width factor in front of the parenthese is too large. 
In order to choose better values for tau, I kept the initial value of the level and plot the curves for different values of tau. I chose values that allow a large hp increase in the lower levels for a fixed amount of food.

mdubied:
**Possible solution to the problem "vertical shift** (see above) @Rhys-J @SaraFFdez  : 
What we could do is have a _single_ curve for the updateHP. We can still enforce the agents to increase their hp by a maximum number of hp on a given day (idea of "recovery" time: it takes time to recover from the weak level to feel strong again). Moreover, we also keep defined level values (critical, weak, healthy, strong). These 4 parameters define the "cost of living": this cost is higher if you are in a higher level. A cost a living is attributed to every 4 levels.
This is a simplification of the model we have, but it seems more reasonable to me. Let me know what you think about this

Rhys-J:
I think simplifying the whole thing is needed. I'm just going to throw out an example to see what people think. 
Things we want the model to do:
- Make it more difficult to maintain hp at higher hp.
- Have agents be able to survive in 'critical state' for multiple days.
- Have a maximum hp gain an agent can get in a given day.

I think this can be achieved by the following:
- In updateHp, keep the critical check (hp<5), but have eating any amount of food put you to 30 hp (I will explain why later)
- Have the value of tau be dependent on your current hp `tau = 5 + ((a.hp-5)*0.333)`
- This gives tau varying between 5 and 28 (for hp from 5 to 75)
- Have a constant width at any hp>5 equal to 40. So in updatehp, the maximum gain is 40hp `a.hp = a.hp + 40- int(math.Pow(math.E, -foodTaken/tau)*float64(40))`
- Have check for hp>100
- In hpDecay, just always decrease the hp by 0.63*40 (=25), unless hp<5 then add to daysInState
- Check for death if in critical state
- This gives the maximum hp gain in a day = 15 hp (40 for eating and -25 for decay)

This basically changes the range of hp you can end a day on to 5 - 75. (excluding critical)
I don't think we need discrete levels of strong, health and weak but if we did then 5-25, 25-50, 50+
If we still wanted to have daysInState count for levels other than critical then we would need to add healthLevel to baseagent struct and check if it has changed at the end of hpDecay.

This will probably need some tweaking but I wanted to make a model that still fits in with the current changes without large overhaul.

pkg/infra/baseagent.go
```diff
@@ -93,8 +95,15 @@ func (a *Base) setHP(newHP int) {
...
 }
 
+// Modeled as a first order system step answer (see documentation for more information)
 func (a *Base) updateHP(foodTaken float64) {
-	a.hp = int(math.Min(100, float64(a.hp)+foodTaken))
+
+	if a.hp >= a.tower.healthInfo.WeakLevel {
+		a.hp = a.hp + int(a.tower.healthInfo.Width*(1-math.Pow(math.E, -foodTaken/a.tower.healthInfo.Tau)))
+	} else {
+		a.hp = int(math.Min(float64(a.tower.healthInfo.HPReqCToW), float64(a.hp+int(a.tower.healthInfo.Width*(1-math.Pow(math.E, -foodTaken/a.tower.healthInfo.Tau))))))
```
jzzheng22, line 104:
```suggestion
        hpChange := a.tower.healthInfo.Width*(1-math.Pow(math.E, -foodTaken/a.tower.healthInfo.Tau))
	if a.hp >= a.tower.healthInfo.WeakLevel {
		a.hp = a.hp + int(hpChange)
	} else {
		a.hp = int(math.Min(float64(a.tower.healthInfo.HPReqCToW), float64(a.hp)+hpChange))
```

jzzheng22 (changes requested):
Will review some more when you're done with the rest of the changes but sending this through now

pkg/infra/baseagent.go
jzzheng22: Actually... should this be max or min? I'm not sure what it's doing.

Rhys-J: `a.hp = a.hp + int(math.Min(float64(a.tower.healthInfo.HPReqCToW), float64(hpChange))`

Not sure if this is what you meant to have here?? This prevents the agent from going into the weak state just to be put back into it. And allows the agent to pass the check in hpDecay of `if agent.hp >= t.healthInfo.HPCritical+t.healthInfo.HPReqCToW` which will then put them into the weak state correctly.

pkg/infra/tower.go
```diff
@@ -98,11 +102,27 @@ func (t *Tower) reshuffle(numOfFloors int) {
...
+				agent.daysAtCritical = 0
+			} else {
+				newHP = t.healthInfo.HPCritical
+				agent.daysAtCritical++
+			}
+		}
+
+		if newHP < t.healthInfo.WeakLevel {
+			newHP = t.healthInfo.HPCritical
+		}
```
jzzheng22, line 122: Can we remove this? I don't think this is ever possible.
```suggestion
```

pkg/infra/tower.go
```diff
@@ -98,11 +102,27 @@ func (t *Tower) reshuffle(numOfFloors int) {
 }
 
 func (t *Tower) hpDecay() {
-	// TODO: can add a parameter
 	for _, agent := range t.agents {
-		newHP := agent.hp - 20
+		newHP := 0
+
+		if agent.hp >= t.healthInfo.WeakLevel {
+			newHP = int(math.Min(float64(t.healthInfo.MaxHP), float64(agent.hp)-(10.0+float64(agent.hp-10)*0.25)))
```
jzzheng22, line 109:
```suggestion
			newHP = int(math.Min(float64(t.healthInfo.MaxHP), float64(agent.hp)-(10.0+float64(agent.hp-t.HealthInfo.WeakLevel)*0.25)))
```

jzzheng22: I'm not sure what the `10.0` here is? Is that also `WeakLevel`?

pkg/utils/globalTypes/health/health.go
```diff
@@ -0,0 +1,29 @@
+package health
+
+type HealthInfo struct {
+	// Maximum HP
+	MaxHP int
+	// Border between critical and weak state
+	WeakLevel int
+	// Parameters of the updateHP function
+	Width float64
+	Tau   float64
```
jzzheng22, line 10: Do we want these as floats or ints?

jzzheng22 (changes requested):
I still feel like HP is more complicated than it needs to be. I don't see why if you're in `critical` then it's always the same HP. I think just decay your HP until zero. I don't think we need to have `maxDaysCritical` either - just decay the HP until you're at 0.

pkg/infra/tower.go
jzzheng22: Never mind, I think I'm wrong

pkg/infra/tower.go
Rhys-J: I think 10 is just a chosen value for the minimum amount of hp you lose

Rhys-J (commented):
If the changes Jason has requested are implemented then I think it all looks good. The main issue is line 104 in baseagent.go. I will re-review with approval once these issues are sorted.

pkg/infra/tower.go
mdubied: It is indeed just a chosen value. We could possibly make it a variable, although I don't think we want it to change if it works correctly

pkg/utils/globalTypes/health/health.go
mdubied: I would rather go for float, as they are only used for computation. That way, we avoid casting operations and longer code. But I think this is a matter of personal preferences

pkg/infra/baseagent.go
mdubied: Rhys is right. There might be other ways to do it but I think it should work quite nicely

pkg/infra/baseagent.go
```diff
@@ -93,8 +95,15 @@ func (a *Base) setHP(newHP int) {
...
 }
 
+// Modeled as a first order system step answer (see documentation for more information)
 func (a *Base) updateHP(foodTaken float64) {
-	a.hp = int(math.Min(100, float64(a.hp)+foodTaken))
+        hpChange := a.tower.healthInfo.Width*(1-math.Pow(math.E, -foodTaken/a.tower.healthInfo.Tau))
+	if a.hp >= a.tower.healthInfo.WeakLevel {
+		a.hp = a.hp + int(hpChange)
+	} else {
+		a.hp = int(math.Min(float64(a.tower.healthInfo.HPReqCToW), float64(a.hp)+hpChange))
```
Rhys-J, line 104: What is it you want this line to do, as your just setting the hp to HPReqCToW (2) each time as this is always going to be the mimimum value??

Rhys-J: Do you want something like `a.hp = a.hp + int(math.Min(float64(a.tower.healthInfo.HPReqCToW), hpChange))` instead?

SaraFFdez (dismissed):
LGTM

pkg/infra/baseagent.go
mdubied: This is a mistake, It should be `math.Min(float64(t.healthInfo.HPCritical+a.tower.healthInfo.HPReqCToW), float64(a.hp)+hpChange)`. Let me commit that

Rhys-J (dismissed):
I think this all looks good now from a functionality perspective. I will leave to Jason to confirm whether readability and code implementation is acceptable. 
Thanks Mathieu for all your work on this!

SaraFFdez (approved):
LGTM

pkg/infra/tower.go
jzzheng22: Happy to leave the 10.0, but the `10` should be changed to `WeakLevel`

pkg/utils/globalTypes/health/health.go
jzzheng22: Okay

jzzheng22 (changes requested):

jzzheng22 (approved):
LGTM, thanks everyone :)

PR #79, opened 2021-12-23
Add ID to message

JaafarRammal:
# Summary

Each message now has a unique ID. If an agent sends out multiple messages, they can now e.g. remember in their state which messages are still pending or which messages have been "resolved".

### DIFF
--- a/pkg/messages/baseMessage.go
+++ b/pkg/messages/baseMessage.go
@@ -1,5 +1,9 @@
 package messages
 
+import (
+	"github.com/google/uuid"
+)
+
 //Define message types to enable basic protocols, voting systems ...etc
 type Message interface {
 	MessageType() string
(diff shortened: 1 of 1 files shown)


pkg/messages/baseMessage.go
```diff
@@ -8,11 +12,13 @@ type Message interface {
 type baseMessage struct {
 	SenderFloor  int
 	responseType Message
+	messageID    string
 }
 
 func NewBaseMessage(SenderFloor int) *baseMessage {
 	msg := &baseMessage{
 		SenderFloor: SenderFloor,
+		messageID:   uuid.New().String(),
```
hussain2603, line 21: The code works fine. Im not too sure how useful having a unique message ID would be. A completely new message ID doesn't act as a conversation ID by which you can 'remember' conversations and who you have talked to. Imo it also doesn't help to remember states about any other agent since it is not associated with an agent and only associated with a message. 

So I personally don't see much use in this, but what does everyone else think?

hussain2603 (changes requested):

jzzheng22 (commented):
Code itself looks fine, but I know that @woonmoon  has been working on messages over the last day or so, so she'll be better placed to review this.

JaafarRammal:
Reply functionality added in #96

PR #90, opened 2021-12-23
Clean logger code

### LINKED ISSUE #36: Clean logger code in main.go
Introduced in #35

jzzheng22:
Closes #36.

This change means we no longer generate `simulation.json`. Because logs are timestamped, to access the latest log we can lexigraphically sort the output logs to determine which log is the most recent one. This also makes the code neater, and means people won't be confused by `simulation.json` and `logs/MOSTRECENTLOG.json`.


JaafarRammal (approved):
A PR that mainly removes code, love to see it! LGTM

PR #95, opened 2021-12-24
Team6 Food Intake Strategy Update

MattScottEEE:
# Summary

Updated food intake strategies based on new HP functions in main

## Additional Information

Agents take food according to social motive and HP function

## Test Plan

No conflicts with infra files - logs have been run to confirm that behaviours are correct

### DIFF
--- a/pkg/agents/team6/strategies.go
+++ b/pkg/agents/team6/strategies.go
@@ -19,8 +19,6 @@ type levelsData struct { // tiers of HP
 	critLevel    float64
 }
 
-var daysInCritical int = 0
-
 func (a *CustomAgent6) foodIntake() food.FoodType {
 	healthInfo := a.HealthInfo()
 
@@ -44,14 +42,12 @@ func (a *CustomAgent6) foodIntake() food.FoodType {
 	case "Collectivist": // Only eat when in critical zone randomly before expiry
 		switch {
 		case currentHP >= levels.weakLevel:
-			daysInCritical = 0
 			a.foodTakeDay = rand.Intn(healthInfo.MaxDayCritical) // Stagger the days when agents return to weak
 			return food.FoodType(0)
 		case currentHP >= levels.critLevel:
-			if daysInCritical == a.foodTakeDay {
+			if a.DaysAtCritical() == a.foodTakeDay {
 				return food.FoodType(healthInfo.HPReqCToW)
 			}
-			daysInCritical++
 			return food.FoodType(0)
 		default:
 			return food.FoodType(0)
--- a/pkg/agents/team6/strategies.go
+++ b/pkg/agents/team6/strategies.go
@@ -20,7 +20,6 @@ type levelsData struct { // tiers of HP
 }
 
 var daysInCritical int = 0
-var foodTakeDay int
 
 func (a *CustomAgent6) foodIntake() food.FoodType {
 	healthInfo := a.HealthInfo()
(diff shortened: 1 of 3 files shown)


pkg/agents/team6/strategies.go
```diff
@@ -1,24 +1,93 @@
...
 
+type levelsData struct {
+	strongLevel  float64
+	healthyLevel float64
+	weakLevel    float64
+	critLevel    float64
+}
+
+var daysInCritical int = 0
+var foodTakeDay int
```
jzzheng22, line 22:
```suggestion
daysInCritical := 0
foodTakeDay := 0
```

pkg/agents/team6/strategies.go
```diff
@@ -1,24 +1,93 @@
...
+
 func (a *CustomAgent6) foodIntake() float64 {
-	thresholds := thresholdData{satisficeThresh: 20.0, satisfyThresh: 60.0, maxIntake: 80.0}
+
```
jzzheng22, line 25:
```suggestion
```

pkg/agents/team6/strategies.go
```diff
@@ -1,24 +1,93 @@
...
+	thresholds := thresholdData{
+		maxIntake: 80.0,
+	}
+
+	levels := levelsData{
+		strongLevel:  0.6 * float64(healthInfo.MaxHP),
+		healthyLevel: 0.3 * float64(healthInfo.MaxHP),
+		weakLevel:    0.1 * float64(healthInfo.MaxHP),
+		critLevel:    0.0,
+	}
```
jzzheng22, line 37: Should this be an int?

pkg/agents/team6/strategies.go
```diff
@@ -1,24 +1,93 @@
...
+	case "Narcissist": // Eat max intake (Possible TODO: Stay in strong instead?)
 		return thresholds.maxIntake
+
 	default:
 		return 0.0
 	}
 }
+
+func foodRequired(currentHP, goalHP float64, healthInfo *health.HealthInfo) float64 {
+	return healthInfo.Tau*math.Log(healthInfo.Width) - healthInfo.Tau*math.Log(healthInfo.Width-goalHP+0.75*currentHP-10+0.25*float64(healthInfo.WeakLevel))
```
jzzheng22, line 80:
```suggestion
        denominator := healthInfo.Width-goalHP+0.75*currentHP-10+0.25*float64(healthInfo.WeakLevel
	return healthInfo.Tau*math.Log(healthInfo.Width / denominator)
```

jzzheng22: Potential simplification? (Check my maths though!)

jzzheng22 (changes requested):

pkg/agents/team6/strategies.go
MattScottEEE: We need these both outside the function, as they get updated inside the function. I think you can only use short assign inside of functions?

pkg/agents/team6/strategies.go
MattScottEEE: Ooh nice nice

pkg/agents/team6/strategies.go
jzzheng22: nonononono no global variables

jzzheng22 (changes requested):
No global variables at all, unless they are consts or singletons. Absolutely no global variables.

pkg/agents/team6/strategies.go
MattScottEEE: Yes, but would require a huge amount of casting both in struct and when accessed in functions

jzzheng22 (approved):

Issue #15, opened 2021-12-09
Move state of agents to simulation environment instead of in tower.


jzzheng22:
Currently a lot of duplicated state which is creating a lot of complexity, e.g.`missingAgents`.

jzzheng22:
Also e.g. `IsDead()` depends on `tower.Exists()`... I think just checking `HP` should work fine here.

PR #101, opened 2021-12-24
Remove duplicated state of agents

### LINKED ISSUE #93: Race condition in `AgentsRun()` in `simIteration.go`
Priority fix - will look into this tomorrow.

In `AgentsRun()` in `pkg/simulation/simIteration.go`, we iterate over the map while writing to it concurrently (see line 47). This is not allowed (https://tip.golang.org/doc/go1.8#mapiter) and is causing a race condition and a crash with the following message:
`fatal error: concurrent map iteration and map write`.

This was identified by @r15hil in Team 4: if you want to see the stack trace please contact him.

I propose the following fix: move the deletion of any dead agents to `sE.TowerTick()`, and also remove the duplication of state. I can't think of any reason for both `simEnv` and `Tower` to contain a list of agents, as this requires both of them to be updated. From what I can tell in the current implementation, the lists are actually inconsistent since they are not updated at the same time.

The `simEnv` map of agents is updated inside `AgentsRun()`, whereas the `Tower` map of agents is updated in `hpDecay()`.

Agents are replaced at the end of each tick, in `replaceAgents()`. The list of agents to replace is obtained from `t.UpdateMissingAgents()`. The `missingAgents` map is updated in `hpDecay()`, with agents added to this map and then removed from the Tower's map of agents.

I think that we can just kill an agent by checking if `agent.IsAlive()` is true or false. This function should be reimplemented to return `a.hp > 0`.

jzzheng22:
# Summary

Closes #93 and #15. 

This PR removes the map of agents from the simulation environment, so that agents are now mapped only in the tower.
We remove the possibility of a race condition (I think) from `AgentsRun()`, as we are no longer writing to `t.Agents` in the goroutine.

Agents are killed inside `tower.Tick()`, in the `hpDecay()` function, by setting their HP to zero.
Dead agents are removed inside `sE.replaceAgents()` by checking to see if `IsAlive()` (which returns `a.hp > 0`) is false. If an agent is dead, it is replaced with a new agent of the same type on that floor.


pkg/infra/tower.go
```diff
@@ -16,9 +16,8 @@ type Tower struct {
 	maxPlatFood    float64
 	currPlatFloor  int
 	agentCount     int
-	agents         map[string]*Base
+	Agents         map[string]Agent
```
udai-arneja, line 19: Why make this public? think its better to have it private.

pkg/infra/tower.go
```diff
@@ -124,16 +117,15 @@ func (t *Tower) hpDecay() {
 		agent.setHasEaten(false)
 		if agent.daysAtCritical >= t.healthInfo.MaxDayCritical {
 			t.Log("Killing agent", Fields{"agent": agent.id})
-			t.missingAgents[agent.floor] = append(t.missingAgents[agent.floor], agent.agentType)
-			delete(t.agents, agent.id) // maybe lock mutex?
-		} else {
-			agent.setHP(newHP)
+			newHP = 0
```
udai-arneja, line 120: why not remove it?

pkg/agents/team1/agent1/agent.go
```diff
@@ -3,15 +3,21 @@ package agent1
...
 	*infra.Base
 	myNumber int
 }
 
-func New(baseAgent *infra.Base) (agent.Agent, error) {
+func New(world world.World, agentType int, agentHP int, agentFloor int, id string) (infra.Agent, error) {
+	baseAgent, err := infra.NewBaseAgent(world, agentType, agentHP, agentFloor, id)
+	if err != nil {
+		log.Fatal(err)
+	}
```
udai-arneja, line 20: Didn't want to put this in the custom agent as it makes it easy for custom agents to manipulate their original health etc.

pkg/agents/randomAgent/agent.go
```diff
@@ -4,16 +4,20 @@ import (
...
 )
 
 type CustomAgentRandom struct {
 	*infra.Base
 	// new params
 }
 
-func New(baseAgent *infra.Base) (agent.Agent, error) {
-	//create other parameters
+func New(world world.World, agentType int, agentHP int, agentFloor int, id string) (infra.Agent, error) {
```
JaafarRammal, line 16: I don't think we should be passing the HP, Floor, and world to custom agents when they initialise a base agent for inheritance. It's counter-intuitive in my opinion because it means someone can just override their agent's starting HP or floor or even their base agent parameters. It also gives custom agents direct access to the world, which exposes the state of the tower, and deprecates the use of a base agent for state abstraction

udai-arneja (changes requested):
IMO makes too many things public, we would then have to write tests to make sure no team changes things etc. Makes it a lot more convoluted, and there isn't massive amounts of  #time.

JaafarRammal (changes requested):
The race condition fix and removing state duplication LGTM, thanks for fixing!

I don't see though the advantage of changing the custom agent creation function in that way given the states it exposes

pkg/infra/tower.go
jzzheng22: Can't access it in simEnv unless it's exported.

pkg/infra/tower.go
jzzheng22: Agents are deleted and replaced in the same place (`replaceAgents()`)

pkg/infra/tower.go
jzzheng22: Also - custom agents can't access it anyway

udai-arneja:
Will have a think for 30mins or so but I like the new changes overall. Thinking about the deleting/replacing happening in the same place (- flexibility wise) and the interface (-for neatness)

r15hil (approved):
This PR seems to have sorted the concurrency issues, code is clean although will require all teams to change their existing "New" functions but not a significant change (a line)

I have also pulled and tested locally trying to recreate the scenario that caused concurrency problem - no errors now

Cheers Jason

udai-arneja (approved):
It's nice, and works. If there are some limitations that come from it later we can always change it. Thanks Jason.

JaafarRammal (approved):

PR #102, opened 2021-12-24
Fixing race issue

udai-arneja:
Removed the deletion to after iteration.

# Summary

Fixing RACE issue - moved the deletion to after iteration.

## Additional Information

Solving issue #93 

## Test Plan

Ran over prolonged period of days. All seems to be clear and making sense

### DIFF
--- a/pkg/infra/baseagent.go
+++ b/pkg/infra/baseagent.go
@@ -37,14 +37,15 @@ func NewBaseAgent(world world.World, agentType int, agentHP int, agentFloor int,
 	}
 	logger := log.WithFields(log.Fields{"agent_id": id, "agent_type": agentType, "reporter": "agent"})
 	return &Base{
-		id:        id,
-		hp:        agentHP,
-		floor:     agentFloor,
-		agentType: agentType,
-		tower:     tower,
-		inbox:     list.New(),
-		logger:    *logger,
-		hasEaten:  false,
+		id:             id,
+		hp:             agentHP,
+		floor:          agentFloor,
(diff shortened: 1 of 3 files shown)


udai-arneja:
@jzzheng22 will wait on ur review before merging X

pkg/simulation/simIteration.go
```diff
@@ -36,19 +36,23 @@ func (sE *SimEnv) TowerTick() {
...
 			}
 			wg.Done()
 		}(&wg, custAgent, uuid)
 	}
 	wg.Wait()
+	sE.mx.Lock()
+	for agentUUID := range agentsToRemove {
+		delete(sE.custAgents, agentsToRemove[agentUUID])
+	}
+	sE.mx.Unlock()
```
jzzheng22, line 57: No use in having a mutex here

pkg/infra/baseagent.go
```diff
@@ -85,8 +85,7 @@ func (a *Base) ID() string {
 }
 
 func (a *Base) IsAlive() bool {
-	_, found := a.tower.agents[a.id]
-	return found
+	return a.hp > 0
```
jzzheng22, line 88: Don't think this will work since a.hp is never set to 0 I think

jzzheng22 (changes requested):

pkg/infra/baseagent.go
udai-arneja: V true hahaa

udai-arneja:
Closing due to better overall solution.

PR #105, opened 2021-12-24
Checking for negative food

### LINKED ISSUE #99: Agents can attempt to take negative food
Agents can attempt to take negative food which puts food back onto the platform and can result in agents getting negative health values.

Should probably have safeguard against this such as checking if amountOfFood > 0 in TakeFood

udai-arneja:
# Summary

Nothing but a hard days work.

Fixes #99


jzzheng22 (approved):
c:

Issue #77, opened 2021-12-22
Implement new information for agents.


SaraFFdez:
I believe the following information would be useful for agent teams:

- When the day has just started
- When a reshuffling has happened
- The tick in which we just ate (or the platform was on our floor but it was empty so we didn't eat but it would have been our turn)

jzzheng22:
1. I'm not 100% convinced if we should do this - how would an agent know if the day has started in the tower?
2. They would know if a reshuffle happened if their floor is not their previous floor (unless by chance they got reshuffled back to the same floor)
3. I'm not sure what you mean by this one... can you explain more please?

SaraFFdez:
1. In the movie the lights turned off at "night time" and turned on again in the morning so I thought an agent would be able to tell what the new day would be or when the day ends. Since this is when the HP decays as well, in our agent implementation this is something we wanted to take into account. (but can do without if we decide we shouldn´t know this)

2. This is how we currently have it implemented yes, by looking at the floor change, but as you say, if the floor is the same, that would be a problem for my group. Not sure how other people are implementing their agents but it might be beneficial for this to be an easier thing to know

3. In our current implementation my team would need to know when we have eaten. Right now we are checking this by seeing if a.TakeFood() returns a number different than 0, but when you don´t eat anything because the platform is empty, this will still return 0 and the agent cannot know that they have tried to eat (unless I am missing something, please let me know). Is there any way we can know that the platform is on our floor and we tried to eat? If not, it would be nice to implement it 

Thoughts?

SaraFFdez:
For number 3, I propose something like changing the last return of the TakeFood function.

```
func (a *Base) TakeFood(amountOfFood float64) float64 {

       if a.floor == a.tower.currPlatFloor && !a.hasEaten {
		foodTaken := math.Min(a.tower.currPlatFood, amountOfFood)
		a.updateHP(foodTaken)
		a.tower.currPlatFood -= foodTaken
		a.setHasEaten(true)
		a.Log("An agent has taken food", Fields{"floor": a.floor, "amount": foodTaken})
		return foodTaken
	}
	return -1.0 <- Return -1 every time we do not eat so we know that if we have attempted to eat but couldn't, we would return 0
} 
```

Idk if this impacts anything else, but if it doesn't I can implement this and change the TakeFood to an int in the process which has not been closed yet

hussain2603:
[...]

PR #112, opened 2021-12-24
Infra improvements

udai-arneja:
# Summary

Food function: -1 if nothing taken, 0 if 0 requested, otherwise amount requested or max food taken.

## Additional Information

Solves issue #77


pkg/infra/baseagent.go
```diff
@@ -142,7 +145,7 @@ func (a *Base) TakeFood(amountOfFood food.FoodType) food.FoodType {
 		a.Log("An agent has taken food", Fields{"floor": a.floor, "amount": foodTaken})
 		return foodTaken
 	}
-	return 0
+	return -1
```
MattScottEEE, line 148: Not sure if this is a good suggestion or not, but might be worth returning an error as well as a quantity?

This would let us differentiate between the cases?

```suggestion
	return 0, errors.New("agent has already eaten/platform isn't on level")
```

pkg/infra/baseagent.go
```diff
@@ -134,6 +134,9 @@ func (a *Base) setHasEaten(newStatus bool) {
 }
 
 func (a *Base) TakeFood(amountOfFood food.FoodType) food.FoodType {
+	if amountOfFood == 0 {
+		return 0
```
MattScottEEE, line 138:
```suggestion
		return 0, errors.New("No food requested to take")
		OR
		return 0, nil
```

udai-arneja:
@SaraFFdez opinons? I wouldn't want to make this more complicated than required. I prefer -1, 0. other x.

pkg/infra/baseagent.go
SaraFFdez: I think -1 is a simple way of signaling that we didn´t eat (instead of 0 which could indicate that we ate 0) As @udai-arneja commented below, I think it's not worth making it more complicated and -1, 0, other x would be easy to handle

SaraFFdez (dismissed):
I think we should keep it simple with -1, 0, x. Otherwise LGTM

pkg/infra/tower.go
```diff
@@ -85,7 +85,7 @@ func (t *Tower) Reshuffle() {
 	// iterate through the uuid strings of each agent
 	for _, agent := range t.Agents {
 		newFloor := rand.Intn(numOfFloors)
-		for remainingVacancies[newFloor] == 0 {
+		for remainingVacancies[newFloor] == 0 && newFloor != agent.BaseAgent().floor {
```
jzzheng22, line 88: Should this be an and or an or?

udai-arneja: apologies, you are correct indeed!

udai-arneja:
Not expecting any reviews till after xmas 😤

SaraFFdez (approved):

Issue #82, opened 2021-12-23
Parametrise backend using config


JaafarRammal:
Currently the backend has parameters in variables. Change that to use a configuration when executing (e.g. using a JSON file)

Issue #81, opened 2021-12-23
Frontend - Backend communication


JaafarRammal:
Setup the communication between the backend and the frontend either via:
- Go executable
- HTTP (server)

Worth checking last year's method for Go executable

PR #115, opened 2021-12-26
Parametrise backend using a config json file

VFjr:
# Summary

This solves issue #82

Main now loads the runtime parameters from a json file, and puts them inside a `ConfigParameters` struct that contains pretty much all parameters needed. 

New `config` package created for the `ConfigParameters` struct and the `LoadParamFromJson` function

The `LoadParamFromJson` function  is used to parse the json and convert it into a struct and is also hopefully very similar to what will be required for the frontend communication to parse the json ( issue #81 )

## Additional Information

Also added functionality for command-line flags to allow the used to choose the location of the config file, however if none is provided, then it will use the config.json file which already contains the default values of previous commit.

### DIFF
--- a/cmd/backend/main.go
+++ b/cmd/backend/main.go
@@ -1,15 +1,15 @@
 package main
 
 import (
+	"flag"
 	"fmt"
 	"math/rand"
 	"os"
 	"path/filepath"
 	"time"
 
+	"github.com/SOMAS2021/SOMAS2021/pkg/config"
 	"github.com/SOMAS2021/SOMAS2021/pkg/simulation"
-	"github.com/SOMAS2021/SOMAS2021/pkg/utils/globalTypes/day"
-	"github.com/SOMAS2021/SOMAS2021/pkg/utils/globalTypes/food"
 	"github.com/SOMAS2021/SOMAS2021/pkg/utils/globalTypes/health"
 	log "github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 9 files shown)


cmd/backend/main.go
```diff
@@ -36,34 +36,21 @@ func main() {
...
+		return
+	}
 
-	healthInfo := health.NewHealthInfo(maxHP, weakLevel, width, tau, hpReqCToW, hpCritical, maxDayCritical, HPLossBase, HPLossSlope)
+	healthInfo := health.NewHealthInfo(parameters.MaxHP, parameters.WeakLevel, parameters.Width, parameters.Tau, parameters.HpReqCToW, parameters.HpCritical, parameters.MaxDayCritical, parameters.HPLossBase, parameters.HPLossSlope)
 
 	// TODO: agentParameters - struct
 
-	simEnv := simulation.NewSimEnv(foodOnPlatform, numOfAgents, agentHP, agentsPerFloor, dayInfo, healthInfo)
+	simEnv := simulation.NewSimEnv(parameters.FoodOnPlatform, parameters.NumOfAgents, parameters.AgentHP, parameters.AgentsPerFloor, parameters.DayInfo, healthInfo)
```
udai-arneja, line 54: I would argue it's worth passing in a pointer to `parameters` to both `simulation.NewSimEnv` and `health.NewHealthInfo` and setting the parameters there. Makes it more readable imo.

config.json
```diff
@@ -0,0 +1,18 @@
+{
+    "FoodOnPlatform": 100,
+    "NumOfAgents": [2,2,2,2,2,2,2],
```
udai-arneja, line 3: maybe making this a little more interpretable, like `"randomAgent": int` and then in the `configparameters.go`, `func LoadParamFromJson` put it into the list in the right places? Makes it a little easier for people to input their agents etc/more readable.

udai-arneja (changes requested):
Looks good overall, thanks for this. Few minor comments.

MattScottEEE:
More of a general comment, but do the command line arguments come with a "-h" parameter, giving a list of the possible arguments? I'm not sure if the flag package handles this implicitly

VFjr:
> More of a general comment, but do the command line arguments come with a "-h" parameter, giving a list of the possible arguments? I'm not sure if the flag package handles this implicitly

Yes, it does that automatically. Also thanks for mentioning it, while testing this I spotted an error in my flag implementation.
image

MattScottEEE (commented):
Overall LGTM, but will let infra team give explicit approval if I've missed something

VFjr:
Oof, tried to rebase with github desktop and looks like I messed something up

JaafarRammal:
> Oof, tried to rebase with github desktop and looks like I messed something up
Why is GitHub desktop a thing, we should all use VIM

VFjr:
I've implemented all of Udai's suggestions, and sorted out the rebase thanks to Jaafar. Never using Github Desktop again.

JaafarRammal (approved):
LGTM thanks Victor.
Please wait for an approval from a team lead who is not on infra team before merging. Thanks!

MattScottEEE (approved):
This all seems good :)

cmd/backend/main.go
VFjr: I've implemented this now, thanks.

config.json
VFjr: I've implemented this now, thanks.

udai-arneja (approved):
Thanks for the changes

PR #117, opened 2021-12-27
Add runtime timeout

JaafarRammal:
Add timeout in case of bugs like deadlocks or infinite loops to not kill the pipeline for 6 hours


.github/workflows/main.yaml
```diff
@@ -61,7 +61,8 @@ jobs:
     
   run:
     name: Run
-    runs-on: ubuntu-latest 
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
```
jzzheng22, line 65:
```suggestion
    timeout-minutes: 15
```

jzzheng22 (changes requested):

.github/workflows/main.yaml
jzzheng22: Actually idk if this is necessary

JaafarRammal: The timeout or the 15?

jzzheng22: The 15

JaafarRammal: Let's put 10 hehe

jzzheng22 (approved):

PR #120, opened 2021-12-28
Logging dead agents during simulation

Rhys-J:
-Added deadAgents mapping parameter to tower struct
-Added UpdateDeadAgents function
-Added DeadAgents function
-Added logs for when agent dies
-Added log for summary of dead agents at end of simulation

# Summary

Added parameter to tower to track dead agents during the simulations.
Added logs for when agent has died and a summary at the end of the simulation.
Added functions to tower.

## Additional Information

Not sure if this is the best way to store and track this information. Please tell me if you have any other suggestions.

Currently works with agentType as int but I understand this is likely to change, so the changes here will have to be modified slightly when this happens.

## Test Plan

Increased the number of simulation days for testing and checked logs over multiple simulations.

### DIFF
--- a/pkg/infra/tower.go
+++ b/pkg/infra/tower.go
@@ -23,6 +23,7 @@ type Tower struct {
 	dayInfo        *day.DayInfo
 	healthInfo     *health.HealthInfo
 	mx             sync.RWMutex
+	deadAgents     map[int]int
 }
 
 func (t *Tower) Log(message string, fields ...Fields) {
@@ -48,6 +49,7 @@ func NewTower(maxPlatFood food.FoodType, agentCount,
 		logger:         *log.WithFields(log.Fields{"reporter": "tower"}),
 		dayInfo:        dayInfo,
 		healthInfo:     healthInfo,
+		deadAgents:     make(map[int]int),
 	}
 }
 
@@ -143,4 +145,12 @@ func (t *Tower) TotalAgents() int {
 	t.mx.RLock()
 	defer t.mx.RUnlock()
 	return len(t.Agents)
+}
+
+func (t *Tower) UpdateDeadAgents(agentType int) {
+	t.deadAgents[agentType]++
+}
+
+func (t *Tower) DeadAgents() map[int]int {
+	return t.deadAgents
 }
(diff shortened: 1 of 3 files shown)


pkg/simulation/simAgent.go
```diff
@@ -48,7 +48,9 @@ func (sE *SimEnv) replaceAgents(t *infra.Tower) {
 		agent := agent.BaseAgent()
 		if !agent.IsAlive() {
 			delete(t.Agents, uuid)
+			sE.Log("An Agent has died", infra.Fields{"Agent Type": agent.AgentType()})
 			sE.createNewAgent(t, agent.AgentType(), agent.Floor())
+			t.UpdateDeadAgents(agent.AgentType())
```
jzzheng22, line 53:
```suggestion
			sE.Log("An Agent has died", infra.Fields{"agent_type": agent.AgentType()})
			t.UpdateDeadAgents(agent.AgentType())
			sE.createNewAgent(t, agent.AgentType(), agent.Floor())

```

pkg/simulation/simulation.go
```diff
@@ -48,6 +48,7 @@ func (sE *SimEnv) Simulate() {
 	sE.Log("Simulation Started")
 	sE.simulationLoop(t)
 	sE.Log("Simulation Ended")
+	sE.Log("Summary of dead agents", infra.Fields{"Agent Type and amount that died": t.DeadAgents()})
```
jzzheng22, line 51:
```suggestion
	sE.Log("Summary of dead agents", infra.Fields{"Agent Type and number that died": t.DeadAgents()})
```

jzzheng22 (changes requested):
Wait for approval from someone in Team 1 as well, not just me

Couple of minor changes, but otherwise looks good.

Upper case vs lower case in logs matters - changes the order in which things are listed since they're sorted alphabetically

jzzheng22 (approved):

pkg/infra/tower.go
```diff
@@ -23,6 +23,7 @@ type Tower struct {
 	dayInfo        *day.DayInfo
 	healthInfo     *health.HealthInfo
 	mx             sync.RWMutex
+	deadAgents     map[int]int
```
JaafarRammal, line 26: I was considering we could have a struct containing al these "states" that we would like to keep track off and log, and then we pass it to tower and sim and base agent. Everyone can then update accordingly the struct and call something like "state.log" which prints the whole state. What do you think ?

JaafarRammal: Actually, that's something we can add on the infra team side when we integrate the frontend logging. I think that should be good for now :)

JaafarRammal (approved):

PR #123, opened 2021-12-28
Infra quickfix

udai-arneja:
# Summary

Max HP an agent can read is 100.

Camel Cased I think


pkg/utils/utilfunctions/utilFunctions.go
```diff
@@ -8,3 +8,10 @@ func Sum(inputList []int) int {
 	return totalAgents
 
 }
+
+func MinInt(valueOne, valueTwo int) int {
+	if valueOne > valueTwo {
+		return valueOne
+	}
+	return valueTwo
+}
```
jzzheng22, line 17:
```suggestion
func MinInt(vars ...int) int {
	min := vars[0]
	for _, i := range vars {
		if min > i {
			min = i
		}
	}
	return min
}
```

jzzheng22 (changes requested):

jzzheng22 (approved):

PR #124, opened 2021-12-28
Change message inbox from list to channel

jzzheng22:
Note that this means we don't need to use a mutex for the inbox


pkg/infra/baseagent.go
```diff
@@ -59,7 +56,7 @@ func NewBaseAgent(world world.World, agentType int, agentHP int, agentFloor int,
 		floor:          agentFloor,
 		agentType:      agentType,
 		tower:          tower,
-		inbox:          list.New(),
+		inbox:          make(chan messages.Message, 15),
```
jzzheng22, line 59: Not sure about the best size to make this buffer - currently set to size 15. This is how many messages an inbox can store before it's full. Can make it larger?

woonmoon (dismissed):
LGTM

woonmoon (approved):

PR #125, opened 2021-12-28
Improve makefile run

JaafarRammal:
# Summary

Call makefile run on the built binary instead of using go
Add arguments to the executable (so we can pass config and other flags)


jzzheng22 (approved):

JaafarRammal:
Will improve with frontend branch

PR #126, opened 2021-12-28
Add code to handle agent updates at end of day

jzzheng22:
# Summary
Following discussion of how we will be handling treaties, I'm changing the code that is run at the end of each day:

- At the end of a day, the function `t.endOfDay()` is called.
- This function then calls:
    - `agent.hpDecay()`
    - `agent.increaseAge()`
    - `agent.updateTreaties()` - this is currently commented out, however once we implement the other treaty code, can be uncommented.
-  There is also an initial implementation, commented out, for `updateTreaties()`.

I've also moved `hpDecay()` from tower to agent, and mildly refactored it to use `utilFunctions.MinInt()` so we don't have to cast the numbers to floats.

I've also added an age field to the base agent - this will be needed when we have treaties.

### DIFF
--- a/pkg/infra/tower.go
+++ b/pkg/infra/tower.go
@@ -66,7 +66,6 @@ func (t *Tower) Tick() {
 	// Decrease agent HP and reset tower at end of day
 	if t.dayInfo.CurrTick%t.dayInfo.TicksPerDay == 0 {
 		t.endOfDay()
-		// t.hpDecay() // decreases HP and kills if < 0
 		t.ResetTower()
 		t.Log("-----------------END----OF----DAY-----------------", Fields{})
 	}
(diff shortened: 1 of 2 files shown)


pkg/infra/tower.go
```diff
@@ -66,7 +65,8 @@ func (t *Tower) Tick() {
 	}
 	// Decrease agent HP and reset tower at end of day
 	if t.dayInfo.CurrTick%t.dayInfo.TicksPerDay == 0 {
-		t.hpDecay() // decreases HP and kills if < 0
+		t.endOfDay()
```
JaafarRammal, line 68: Hmm I'd say let's have the treaty part in its own PR once the implementation is done and keep this nice `t.endOfDay()` change?

JaafarRammal (approved):
thanks

PR #137, opened 2021-12-29
Infra/treaty messages

woonmoon:
Part of on-going treaty building PRs.


pkg/messages/baseMessage.go
```diff
@@ -55,6 +57,12 @@ type RequestMessage interface {
...
 }
 
+type ProposalMessage interface {
+	Message
+	Treaty() Treaty
+	Reply(senderFloor int, response bool) ResponseMessage
+}
+
 type ResponseMessage interface {
 	Message
```
MattScottEEE, line 67:
```suggestion
	Message
	Treaty() Treaty
```

Just for when baseAgent sticks this stuff into the memory, I'm pretty sure we'll have to extract the treaty from the response

pkg/messages/treatyMessage.go
```diff
@@ -0,0 +1,116 @@
...
+
+func NewTreaty(condition ConditionType, request RequestType, cop Op, rop Op, duration int) *Treaty {
+	treaty := &Treaty{
+		condition:      condition,
+		request:        request,
+		conditionOp:    cop,
+		requestOp:      rop,
+		signatureCount: 0,
+		duration:       duration,
+		id:             uuid.New().String(),
```
jzzheng22, line 59: Can we leave it as `uuid.New()` instead of then changing it to string? Would prefer the type safety if possible.

pkg/messages/baseMessage.go
```diff
@@ -55,6 +57,12 @@ type RequestMessage interface {
 	Reply(senderFloor int, response bool) ResponseMessage
 }
 
+type ProposalMessage interface {
+	Message
+	Treaty() Treaty
+	Reply(senderFloor int, response bool) ResponseMessage
```
MattScottEEE, line 63:
```suggestion
```

Do we need a response if we're just making a proposal?

pkg/messages/treatyMessage.go
```diff
@@ -0,0 +1,116 @@
...
+
+const (
+	GT Op = iota + 1
+	GE
+	EQ
+	LE
+	LT
+)
+
+type Treaty struct {
```
MattScottEEE, line 31: Said in the comments (and I could be wrong on this) but I think this should be in utils and imported, as agents will have to import this data type. Ik we import messages but I'm annoying

jzzheng22 (changes requested):

pkg/messages/treatyMessage.go
```diff
@@ -0,0 +1,116 @@
...
+	Id() string
+}
+
+func NewTreaty(condition ConditionType, request RequestType, cop Op, rop Op, duration int) *Treaty {
+	treaty := &Treaty{
+		condition:      condition,
+		request:        request,
+		conditionOp:    cop,
+		requestOp:      rop,
+		signatureCount: 0,
```
MattScottEEE, line 57:
```suggestion
		signatureCount: 1,
```

If we're proposing a treaty, we've implicitly already signed it

MattScottEEE (changes requested):
Looks really good, bar a few minor things. Need others' opinions if this is useful or not

pkg/messages/baseMessage.go
jzzheng22: Yes...?

pkg/messages/treatyMessage.go
jzzheng22: Ehh personally I don't see the point, is there a use case you can think of for why this is better?

MattScottEEE: Mainly convention, since we have healthInfo as a struct in a separate file and also bc I'd rather use treaties.Treaty instead of messages.Treaty. Shouldn't be mixing the 2 up

pkg/messages/baseMessage.go
MattScottEEE: But it's a proposal - we're not replying to anything

pkg/messages/baseMessage.go
jzzheng22: I disagree with adding `Treaty` to the response message... but then this has brought up the question of: how does the recipient agent know what the responseMessage is responding to?

MattScottEEE: Nah I think I'll need it for the memory pushing - I need to handle a response and extract the treaty from there. I don't want to find the message this replies to

pkg/messages/baseMessage.go
jzzheng22: The recipient of `ProposalMessage` calls `Reply()` to respond to the proposal

MattScottEEE: Ah ok, my lack of understanding. I agree then

pkg/messages/baseMessage.go
jzzheng22: I understand that, but multiple message types use `ResponseMessage` as a reply, not just `ProposalMessage`. Therefore, it shouldn't have `Treaty`.

MattScottEEE: Then we need a new dataType to handle this

MattScottEEE: Proposal message will need to have the reply of type TreatyResponseMessage that contains the original treaty

jzzheng22: I'd disagree with that - it would, I think, be better for agents to keep track of what messages they've sent (and therefore what treaties they've proposed). This allows `ResponseMessage()` to stay general.

MattScottEEE: How about containing a pointer to the original message inside of here

MattScottEEE (changes requested):

hussain2603 (commented):
LGTM, I have some of the same suggestions as above so just need to address them.

pkg/messages/treatyMessage.go
JaafarRammal: I thought we had agreed that a treaty was a substitute of messages since you can send a receive a treaty? (it's basically a message on steroids)

If that's the case, I think it's best to keep it as is in the messages package

pkg/messages/treatyMessage.go
JaafarRammal: I think we have to change this in a few other places

jzzheng22: Yup, I created an issue for it a while back.

pkg/messages/treatyMessage.go
MattScottEEE: I just think treaties are used outside of messages too much to be kept in here

pkg/messages/treatyMessage.go
```diff
@@ -0,0 +1,116 @@
...
+	treaty *Treaty
+}
+
+func NewProposalMessage(senderFloor int, condition ConditionType, request RequestType, cop Op, rop Op, duration int) *ProposeTreatyMessage {
+	msg := &ProposeTreatyMessage{
+		NewBaseMessage(senderFloor, ProposeTreaty),
+		NewTreaty(condition, request, cop, rop, duration),
+	}
+	return msg
+}
```
jzzheng22, line 103: I think it's better to pass in a treaty rather than pass in the parameters of the treaty here

pkg/messages/treatyMessage.go
```diff
@@ -0,0 +1,116 @@
...
+	return t.signatureCount
+}
+
+func (t *Treaty) Duration() int {
+	return t.duration
+}
+
+func (t *Treaty) Id() uuid.UUID {
+	return t.id
+}
```
jzzheng22, line 90: This should probably go into a separate file `treaty.go`

pkg/messages/treatyMessage.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+type Treaty struct {
+	condition      ConditionType
+	request        RequestType
+	conditionOp    Op
+	requestOp      Op
+	signatureCount int
+	duration       int
+	id             uuid.UUID
```
jzzheng22, line 38: Should these fields be exported or unexported?

At the very least, surely `signatureCount` should be exported? Or a `signTreaty()` function is needed?

jzzheng22 (changes requested):

pkg/messages/treatyMessage.go
MattScottEEE: this is what I've been saying lol

jzzheng22: I think it should be a different file within the messages package, not in a different package.

pkg/messages/treatyMessage.go
woonmoon: Unexported I think and Ill add `signTreaty()`

pkg/messages/treatyResponseMessage.go
```diff
@@ -0,0 +1,28 @@
...
+	return msg.response
+}
+
+func (msg *TreatyResponseMessage) Treaty() *Treaty {
+	return msg.treaty
+}
+
+func (msg *TreatyResponseMessage) Visit(a Agent) {
+	a.HandleTreatyResponse(*msg)
+}
```
jzzheng22, line 28: I still don't see the point of this - why not just use ResponseMessage??

pkg/messages/proposeTreatyMessage.go
```diff
@@ -0,0 +1,27 @@
+package messages
+
+type ProposeTreatyMessage struct {
+	*BaseMessage
+	treaty *Treaty
```
jzzheng22:
```suggestion
	treaty Treaty
```

pkg/messages/proposeTreatyMessage.go
```diff
@@ -0,0 +1,27 @@
...
+}
+
+func NewProposalMessage(senderFloor int, treaty *Treaty) *ProposeTreatyMessage {
```
jzzheng22:
```suggestion
func NewProposalMessage(senderFloor int, treaty Treaty) *ProposeTreatyMessage {
```

jzzheng22 (approved):

jzzheng22 (approved):

pkg/messages/treaty.go
```diff
@@ -0,0 +1,95 @@
...
+const (
+	GT Op = iota + 1
+	GE
+	EQ
+	LE
+	LT
+)
+
+type Treaty struct {
+	condition      ConditionType
```
MattScottEEE, line 32:
```suggestion
        proposerID   uuid.UUID
        responderID uuid.UUID
	condition      ConditionType
```

Would need these to keep track of who the treaty is between for opinion formation

MattScottEEE (dismissed):
Just this small change to the struct. Otherwise LGTM :)

pkg/messages/treaty.go
MattScottEEE: sorry for the formatting eeeew

woonmoon: Done!

pkg/messages/baseMessage.go
```diff
@@ -58,6 +61,12 @@ type RequestMessage interface {
 	Reply(senderID uuid.UUID, senderFloor int, response bool) ResponseMessage
 }
 
+type ProposalMessage interface {
+	Message
+	Treaty() Treaty
+	Reply(senderFloor int, response bool) ResponseMessage
```
jzzheng22, line 67:
```suggestion
	Reply(senderID uuid.UUID, senderFloor int, response bool) ResponseMessage
```

jzzheng22 (changes requested):

jzzheng22 (approved):

jzzheng22 (approved):

PR #139, opened 2021-12-29
Team5 maxhp

tchiko7:
# Summary

added function to get max  hp and percentage hp

## Additional Information

also changed existing hp getters to call the percentageHP function

### DIFF
--- a/pkg/agents/team5/agent1/agent.go
+++ b/pkg/agents/team5/agent1/agent.go
@@ -67,22 +67,22 @@ func (a *CustomAgent5) updateAim() {
 	case a.selfishness >= 3:
 		// If fully selfish always try to gain health
 		a.currentAim = 2
-	case PercentageHP(a) > 0.80 && a.selfishness == 2:
+	case PercentageHP(a) > 80 && a.selfishness == 2:
 		// Try to maintain health if near max health if mostly selfish
 		a.currentAim = 1
-	case PercentageHP(a) > 0.80 && a.selfishness < 2:
+	case PercentageHP(a) > 80 && a.selfishness < 2:
 		// Willing to lose health near max health if mostly or completely selfless
 		a.currentAim = 0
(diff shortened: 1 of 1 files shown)


hussain2603 (approved):

pkg/agents/team5/agent1/agent.go
```diff
@@ -49,6 +49,14 @@ func New(baseAgent *infra.Base) (infra.Agent, error) {
 	}, nil
 }
 
+func MaxHP(a *CustomAgent5) int {
+	return a.HealthInfo().MaxHP
+}
+
+func PercentageHP(a *CustomAgent5) float64 {
+	return float64(a.HP()) / float64(MaxHP(a))
+}
```
jzzheng22, line 58: I'd prefer to keep this as an int

pkg/agents/team5/agent1/agent.go
```diff
@@ -49,6 +49,14 @@ func New(baseAgent *infra.Base) (infra.Agent, error) {
 	}, nil
 }
 
+func MaxHP(a *CustomAgent5) int {
+	return a.HealthInfo().MaxHP
+}
```
jzzheng22, line 54: Why not just call `a.HealthInfo().MaxHP` directly??

jzzheng22 (changes requested):

pkg/agents/team5/agent1/agent.go
tchiko7: i thought code would be easier to read, i can change if you like

jzzheng22: Ehhh we only use it in one place so far, I think we don't need the function - if we use a lot more and it gets annoying, then maybe yeah.

jzzheng22 (approved):

jzzheng22:
Thanks Osman

PR #140, opened 2021-12-29
Add UUIDs to messages

hussain2603:
# Summary

If a message is of type ResponseMessage, then the requestMessage id is added to the struct and can be accessed through the ReturnId() function.


pkg/messages/baseMessage.go
```diff
@@ -63,12 +68,17 @@ type ResponseMessage interface {
 type BaseMessage struct {
 	senderFloor int
 	messageType MessageType
+	id          string
 }
 
-func NewBaseMessage(senderFloor int, messageType MessageType) *BaseMessage {
+func NewBaseMessage(senderFloor int, messageType MessageType, id string) *BaseMessage {
```
JaafarRammal, line 74: Isn't there a way to make this id optional and it fills itself to empty by default? (instead of having all new messages forced to add empty ID)

jzzheng22: Yes there is. And also... not how I would do this anyway

jzzheng22 (changes requested):
Shouldn't be including such major changes to your individual agent files in this PR.

pkg/messages/baseMessage.go
JaafarRammal: Nevermind lol, Go does not support optional arguments nor function overloading... I guess this is the way...

JaafarRammal (approved):
LGTM
My only comment would be about having to pass an empty ID instead of being able to not pass anything

JaafarRammal (changes requested):
Oops turns our there is a way to make this empty string not a thing in the input... if we could have this it would be fab 😄

pkg/messages/baseMessage.go
jzzheng22: @JaafarRammal  @hussain2603  you can use variadic arguments https://stackoverflow.com/a/19813113

JaafarRammal:
Also looking further, I have a general comment: I think this feature is basically conversations, and in my opinion architecturally the messages shouldn't have the same ID, but rather have unique IDs and somehow be related to a unique conversation.

I also think it's enough if a message has a unique ID and you reply back to a message with the intended message ID you are replying to

jzzheng22:
Have just called them, proposed changes: each message has its own UUID. Response messages will have their own UUID + contain the UUID of the message they are responding to. Suggested the use of variadic arguments where relevant. Also see #79.

pkg/agents/team1/agent1/agent.go
```diff
@@ -51,10 +51,10 @@ func (a *CustomAgent1) Run() {
...
 		a.SendMessage(1, msg)
 		a.Log("I sent a message", infra.Fields{"message": "RequestTakeFood"})
-	case 5:
-		msg := messages.NewResponseMessage(a.Floor(), true)
-		a.SendMessage(1, msg)
-		a.Log("I sent a message", infra.Fields{"message": "Response"})
+	// case 5:
+	// 	msg := messages.NewResponseMessage(a.Floor(), true, )
+	// 	a.SendMessage(1, msg)
+	// 	a.Log("I sent a message", infra.Fields{"message": "Response"})
```
jzzheng22, line 57: Why is this commented out?

pkg/agents/team6/config.go
```diff
@@ -97,7 +97,7 @@ func (a *CustomAgent6) Run() {
 	a.Log("Team 6 took:", infra.Fields{"foodTaken": foodAmount, "bType": a.currBehaviour.String()})
 	a.Log("Team 6 agent has HP:", infra.Fields{"hp": a.HP()})
 
-	msg := messages.NewResponseMessage(a.Floor(), true)
+	msg := messages.NewAskHPMessage(a.Floor())
```
jzzheng22, line 100: Why are you changing this???

jzzheng22 (changes requested):

pkg/agents/team6/config.go
hussain2603: because responseMessage can not be sent without a requestMessage.

pkg/agents/team1/agent1/agent.go
hussain2603: because responseMessage can not be sent without a requestMessage.

pkg/agents/team6/config.go
jzzheng22: ...that's a good point

pkg/agents/team1/agent1/agent.go
jzzheng22: Can you add in a default case handler? Otherwise this is going to error.

pkg/agents/team6/config.go
jzzheng22: Pinging @MattScottEEE  since this is your team's code.

pkg/agents/team1/agent1/agent.go
```diff
@@ -52,7 +52,7 @@ func (a *CustomAgent1) Run() {
 		a.SendMessage(1, msg)
 		a.Log("I sent a message", infra.Fields{"message": "RequestTakeFood"})
 	case 5:
-		msg := messages.NewResponseMessage(a.Floor(), true)
+		msg := messages.NewAskHPMessage(a.Floor())
```
jzzheng22, line 55: Just delete the entire case if you're going to do this

jzzheng22 (changes requested):

pkg/agents/team1/agent1/agent.go
JeetendraJoshi: Fixed this so the case can still run but nothing will happen.

jzzheng22: Just delete it??

pkg/messages/baseMessage.go
```diff
@@ -52,23 +57,26 @@ type StateMessage interface {
 type RequestMessage interface {
 	Message
 	Request() int
-	Reply(senderFloor int, response bool) ResponseMessage
+	Reply(senderFloor int, response bool, returnId uuid.UUID) ResponseMessage
```
JaafarRammal, line 60: I'm finding it hard to get my head around this, could someone explain? Is this basically saying "a message of type request message has to implement a plausible response message that can be used to reply back"? If so, shouldn't that just be a message type? Or would you do `Message.Reply`? I'm a bit confused what's the logic here

jzzheng22: I think the variable name `returnId` could be improved... just not sure what for.

The `Reply()` function is called by the agent which receives the `RequestMessage()`, and this then creates the reply message. `returnId` is the id of the message which the agent is replying to, i.e. the `RequestMessage`s UUID.

jzzheng22: (I think)

JaafarRammal: I'm still confused, sorry. So the agent receives a message `receivedMessage` in their handler, what is `receivedMessage.Reply(...)` doing here? Does it create a response message that I can then send? If so does it mean it can only create a response message of type bool?

JaafarRammal: so the agent would send back `receivedMessage.Replay(a.Floor(), RESPONSE, ?????)`? what's `returnId` here?

JaafarRammal: I think I get it. I think there is duplication here. You are already calling `Reply` on the message, and then you are also passing in the `returnId` and using it. Shouldn't it be either? Since you're calling `Reply` on the message you should be able to pull the message id directly (the whole point of this convo implementation is to reduce implementation complexity so agent doesn't need to be aware of both the ID AND the reply mechanism)

jzzheng22, line 60:
```suggestion
	Reply(senderFloor int, response bool) ResponseMessage
```

jzzheng22: You don't need to pass the UUID in the reply function here. Also, doing so would be wrong anyway.

JaafarRammal: Resolved in private convo, thanks Jason

pkg/messages/baseMessage.go
```diff
@@ -52,23 +57,26 @@ type StateMessage interface {
...
 }
 
 type ResponseMessage interface {
 	Message
 	Response() bool
+	ReturnId() uuid.UUID
```
jzzheng22, line 66:
```suggestion
	RequestId() uuid.UUID
```

pkg/messages/responseMessage.go
```diff
@@ -1,14 +1,20 @@
 package messages
 
+import (
+	"github.com/google/uuid"
+)
+
 type BoolResponseMessage struct {
 	*BaseMessage
 	response bool
+	returnId uuid.UUID
```
jzzheng22, line 10:
```suggestion
	requestId uuid.UUID
```

pkg/messages/responseMessage.go
```diff
@@ -1,14 +1,20 @@
...
 }
 
-func NewResponseMessage(senderFloor int, response bool) *BoolResponseMessage {
+func NewResponseMessage(senderFloor int, response bool, returnId uuid.UUID) *BoolResponseMessage {
 	msg := &BoolResponseMessage{
 		NewBaseMessage(senderFloor, Response),
 		response,
+		returnId,
```
jzzheng22, line 17:
```suggestion
func NewResponseMessage(senderFloor int, response bool, requestId uuid.UUID) *BoolResponseMessage {
	msg := &BoolResponseMessage{
		NewBaseMessage(senderFloor, Response),
		response,
		requestId,
```

pkg/messages/responseMessage.go
```diff
@@ -17,6 +23,10 @@ func (msg *BoolResponseMessage) Response() bool {
 	return msg.response
 }
 
+func (msg *BoolResponseMessage) ReturnId() uuid.UUID {
+	return msg.returnId
+}
```
jzzheng22, line 28:
```suggestion
func (msg *BoolResponseMessage) RequestId() uuid.UUID {
	return msg.requestId
}
```

jzzheng22 (changes requested):

pkg/agents/team6/config.go
MattScottEEE: Totally fine - we were just playing around with message types so this code doesn't convey any real meaning outside of MVP

pkg/agents/team1/agent1/agent.go
MattScottEEE: Same as team6 - I presume this was template stuff?

in which case, this is fine to either delete/replace

MattScottEEE (commented):
In general good to me, just need to work out if passing a pointer to the sender message is better/worse than the UUID. Team6 stuff is fine otherwise, I presume Jason and I will hash this out by tomorrow

JaafarRammal (dismissed):
Apologies for no getting it early on, looks good!

jzzheng22 (approved):
Thanks!

PR #145, opened 2021-12-29
Refactor updateHP

jzzheng22:
# Summary

### DIFF
--- a/pkg/infra/baseagent.go
+++ b/pkg/infra/baseagent.go
@@ -135,11 +135,11 @@ func (a *Base) setHP(newHP int) {
 
 // Modeled as a first order system step answer (see documentation for more information)
 func (a *Base) updateHP(foodTaken food.FoodType) {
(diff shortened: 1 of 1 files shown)


pkg/infra/baseagent.go
```diff
@@ -135,11 +135,11 @@ func (a *Base) setHP(newHP int) {
 
 // Modeled as a first order system step answer (see documentation for more information)
 func (a *Base) updateHP(foodTaken food.FoodType) {
-	hpChange := a.tower.healthInfo.Width * (1 - math.Pow(math.E, -float64(foodTaken)/a.tower.healthInfo.Tau))
+	hpChange := int(a.tower.healthInfo.Width * (1 - math.Pow(math.E, -float64(foodTaken)/a.tower.healthInfo.Tau)))
```
JaafarRammal, line 138: Why is this being set as an Int?

jzzheng22: Because the expression returns a float64, so needs to be cast to int.

jzzheng22: It's not needed as a float anywhere

JaafarRammal (approved):

PR #148, opened 2021-12-29
Team2

tomeaton17:
# Summary

Adds the MVP of Team 2. Amount of food taken by agent depends on the status of itself and other agents.
## Additional Information
More information about agent strategy can be found in `pkg/agents/team2/config.go`

## Test Plan
Tested by running simulation with our agent.


pkg/agents/team2/actions.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	initialActionSpace.actionSet = m
+	return initialActionSpace
+}
+
+//Need to change this func when adding new actions
+
+func DisFood(hp int) food.FoodType {
+	return 0
+}
```
jzzheng22, line 33: What does this function actually do?

pkg/agents/team2/actions.go
```diff
@@ -0,0 +1,64 @@
...
+
+func Satisfice(hp int) food.FoodType {
+	if hp <= 20 { //20 is the daily decay
```
jzzheng22, line 36: This is now no longer true: see https://github.com/SOMAS2021/SOMAS2021/wiki/Health for documentation.
In summary, your HP decay is dependent on your current hp.

pkg/agents/team2/actions.go
```diff
@@ -0,0 +1,64 @@
...
+		return 20
+	} else {
+		return 1
+	}
```
jzzheng22, line 40:
```suggestion
	}
	return 1
```

pkg/agents/team2/actions.go
```diff
@@ -0,0 +1,64 @@
...
+}
+func Satisfy(hp int) food.FoodType {
+
```
jzzheng22, line 43:
```suggestion
```

pkg/agents/team2/config.go
```diff
@@ -0,0 +1,126 @@
...
+	//Communication & Observation
+	//communicate before platform arrives to my floor
+
+	/*
+		msg := *messages.NewBaseMessage(a.Floor())
+		a.SendMessage(1, msg)
+	*/
+
+	//Perform the following only when platform arrives
+	//NOTE: should let infra team add a func to see whether the plaftfrom has arrived or not
```
jzzheng22, line 106: Just as an FYI, you could possibly use `baseagent.CurrPlatFood()` as a proxy for this?

pkg/agents/team2/config.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	if a.isPlatArrived {
+		oldState := a.CheckState()
+		oldHP := a.HP()
+		a.Log("Agent team2 before action:", infra.Fields{"floor": a.Floor(), "hp": oldHP, "food": a.CurrPlatFood(), "state": oldState})
+		action := a.SelectAction()
+		a.TakeFood(a.actionSpace.actionSet[action](oldHP)) //perform selected action
+		a.Log("Agent team2:", infra.Fields{"selected and performed action": action})
+		a.Log("Agent team2 after action:", infra.Fields{"floor": a.Floor(), "hp": a.HP(), "food": a.CurrPlatFood(), "state": a.CheckState()})
+		hpInc := a.HP() - oldHP
```
jzzheng22, line 120: FYI, your actual hp at the start of the next tick would be `a.HP() - some decay`

pkg/agents/team2/policies.go
```diff
@@ -0,0 +1,43 @@
+package team2
+
+func InitPolicies(numStates int, numActions int) [][]float32 {
+	var policies = make([][]float32, numStates)
```
jzzheng22, line 4:
```suggestion
	policies := make([][]float32, numStates)
```

pkg/agents/team2/policies.go
```diff
@@ -0,0 +1,43 @@
...
+	for i := 0; i < numStates; i++ {
+		policies[i] = make([]float32, numActions)
+		for j := 0; j < 3; j++ {
+			policies[i][j] = uniformProb
+		}
+	}
+	return policies
+}
+
+func (a *CustomAgent2) updataPolicies(state int) {
```
jzzheng22, line 15:
```suggestion
func (a *CustomAgent2) updatePolicies(state int) {
```

jzzheng22: Typo...?

pkg/agents/team2/policies.go
```diff
@@ -0,0 +1,43 @@
...
+	for _, policy := range a.policies {
+		sum := float32(0.0)
+		for _, prob := range policy {
+			sum += prob
+		}
+		if sum-1.0 != 0.0 {
+			policy[0] -= sum - 1.0
+		}
+	}
+}
```
jzzheng22, line 43: What's the point of this function?

pkg/agents/team2/qtable.go
```diff
@@ -0,0 +1,42 @@
...
+	maxQ       float32
+}
+
+func max(a float32, b float32) float32 {
+	if a >= b {
+		return a
+	} else {
+		return b
+	}
+}
```
jzzheng22, line 14: If you use `float64` (which you probably should) then you can use `math.Max()`

pkg/agents/team2/reward.go
```diff
@@ -0,0 +1,29 @@
...
+func calcReward(hp int, hpInc int) float32 {
+	//TODO: reward should be redone later according to new healthInfo
+	ret := float32(0.0)
+
+	//we encourage agent to survive
+	if hp > 20 {
+		ret += 1.0
+	} else {
+		ret -= 0.5
+	}
```
jzzheng22, line 12: I'm assuming this was based on the assumption that hp decays by 20 each day.

pkg/agents/team2/reward.go
```diff
@@ -0,0 +1,29 @@
...
+	//we encourage ageny to eat less when hp level is high
+	oldHP := float32(hp - hpInc)
+	incRate := float32(hpInc) / oldHP
+	if incRate > 1.0 {
+		ret += 1.0
+	} else {
+		ret -= 1.0
+	}
+	return ret
```
jzzheng22, line 21:
```suggestion
	if incRate > 1.0 {
		return ret + 1.0
	}
	return ret - 1.0
```

pkg/agents/team2/states.go
```diff
@@ -0,0 +1,36 @@
+package team2
+
+func InitStateSpace(dim1 int, dim2 int, dim3 int) [][][]int {
+	var stateSpace = make([][][]int, dim1) // x axis
```
jzzheng22, line 4:
```suggestion
	stateSpace := make([][][]int, dim1) // x axis
```

jzzheng22 (changes requested):
Is there a reason you're using `float32` instead of `float64`? I think it better to use `float64` by default.

Also, I think you'll have to edit `config.json` as well to add your agent there too.

pkg/agents/team2/config.go
JaafarRammal: I think there is a function for that now available right? cc @MattScottEEE

pkg/agents/team2/config.go
```diff
@@ -0,0 +1,126 @@
...
+		a.updateRTable(hpInc, oldState, action)
+		a.updateQTable(oldState, action)
+		a.updataPolicies(oldState)
```
JaafarRammal, line 123:
```suggestion
		a.updatePolicies(oldState)
```
(if you fix other typo instance)

pkg/agents/team2/actions.go
laiwenq: For our MVP, the logic of this function is not implemented yet. Now it always returns zero as the amount of food to be taken.

jzzheng22: I assumed so, would be good if you could document the intended behaviour for future reference or at least mark it as a todo

pkg/agents/team2/actions.go
laiwenq: Yes, this was implemented before the introduction of Health. please ignore the comment for now as it still makes sense for an MVP. The logic will be substituted in the subsequent versions i.e. using the newly introduced health level instead of hp. Yes, again, I should've added TODO.

pkg/agents/team2/actions.go
laiwenq: will do.

pkg/agents/team2/config.go
laiwenq: I did it. the CheckState function uses CurrPlatFood to determine whether the platform has arrived or not.

pkg/agents/team2/policies.go
laiwenq: To deal with potential rounding errors caused by updating probabilities. Like after several updates, the sum of probability could exceed 1.0, which can crash the program.

pkg/agents/team2/qtable.go
laiwenq: Used float32 to reduce memory consumption, as the tables will expand exponentially in future versions.

pkg/agents/team2/reward.go
laiwenq: Yes, the logic has to be rewritten for next version (together with message passing).

pkg/agents/team2/reward.go
laiwenq: will do.

pkg/agents/team2/states.go
laiwenq: Oops.

laiwenq:
> 

I think Team2Agents it is already in the config.json file?

pkg/agents/team2/config.go
laiwenq: The current logic is: 1) Get hp at the start of Run() and store it as OldHP; 2) Select an action based on OldHP; 3) Perform the action; 4) Get update HP and calculate the increment.
The HP decay happens after the end of day, isn't it? I don't understand why the decay matters here.

pkg/agents/team2/reward.go
laiwenq: No, actually the Reward calculation will get more sophisticated, and it is better to leave the return value at the end line. Gonna change it, but might bring it back later.

pkg/agents/team2/qtable.go
JaafarRammal: I agree reducing the type can reducing memory consumption, just how big is the table? a few bytes or way more?

pkg/agents/team2/policies.go
laiwenq: fixed

pkg/agents/team2/policies.go
JaafarRammal: I haven't reviewed the PR so I can't comment about the implementation, but `crash the program` scares me here 😅
Things that can crash the program are e.g. divisions by 0 or illegal memory accesses. Are we talking about things like that here?

pkg/agents/team2/qtable.go
laiwenq: Depends on how well we want the agent to learn. The current plan is to limit it to 1000 elements per table. If that is not enough we might opt for neural networks that can replace tabular methods.

JaafarRammal: I think using either wouldn't matter because it's a x2. If the scaling up is exponential, there will be a memory problem regardless of the type after a certain scale degree.

That being said, keeping `float32` isn't really an issue imo

pkg/agents/team2/policies.go
laiwenq: Illegal memory access might be induced when the probability exceeds 1.0. For the current implementation, this should never happen. This function is to provide extra safety.

JaafarRammal: Cool, thanks for confirming

hussain2603 (changes requested):
I think you will need to change the config.json file as well since you are now pushing out the randomAgent.

pkg/agents/team2/policies.go
```diff
@@ -0,0 +1,43 @@
...
+		policies[i] = make([]float64, numActions)
+		for j := 0; j < 3; j++ {
+			policies[i][j] = uniformProb
+		}
+	}
+	return policies
+}
+
+func (a *CustomAgent2) updatePolicies(state int) {
+	Delta := float64(0.1) / float64(len(a.actionSpace.actionId)-1)
```
jzzheng22, line 16:
```suggestion
	Delta := 0.1 / float64(len(a.actionSpace.actionId)-1)
```

pkg/agents/team2/policies.go
```diff
@@ -0,0 +1,43 @@
...
+	bestAction := a.getMaxQ(state).bestAction
+	sum := float64(0.0)
```
jzzheng22, line 18:
```suggestion
	sum := 0.0
```

pkg/agents/team2/config.go
```diff
@@ -0,0 +1,126 @@
...
+	stateSpace    [][][]int
+	actionSpace   actionSpace
+	policies      [][]float64
+	rTable        [][]float64
+	qTable        [][]float64
+	isPlatArrived bool
+}
+
+func InitTable(numStates int, numActions int) [][]float64 {
+	var table = make([][]float64, numStates)
```
jzzheng22, line 72:
```suggestion
	table := make([][]float64, numStates)
```

pkg/agents/team2/policies.go
```diff
@@ -0,0 +1,43 @@
...
+		}
+	}
+	a.policies[state][bestAction] = 1.0 - sum
+	a.adjustPolicies()
+}
+
+//fix small errors caused during policy update
+func (a *CustomAgent2) adjustPolicies() {
+	for _, policy := range a.policies {
+		sum := float64(0.0)
```
jzzheng22, line 35:
```suggestion
		sum := 0.0
```

pkg/agents/team2/qtable.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+	return ret
+}
+func (a *CustomAgent2) updateQTable(state int, action int) {
+	//TODO: take hyperparameters out of function
+
+	// Gamma is the discount factor (0≤γ≤1). It determines how much importance we want to give to future
+	// rewards. A high value for the discount factor (close to 1) captures the long-term effective award, whereas,
+	// a discount factor of 0 makes our agent consider only immediate reward, hence making it greedy.
+	Gamma := float64(0.6)
```
jzzheng22, line 32:
```suggestion
	Gamma := 0.6
```

pkg/agents/team2/qtable.go
```diff
@@ -0,0 +1,38 @@
...
+
+	// Alpha is the learning rate (0<α≤1). Alpha is the extent to which our Q-values are being updated in every iteration.
+	Alpha := float64(0.1)
```
jzzheng22, line 35:
```suggestion
	Alpha := 0.1
```

pkg/agents/team2/reward.go
```diff
@@ -0,0 +1,37 @@
+package team2
+
+func calcReward(hp int, hpInc int) float64 {
+	//TODO: reward should be redone later according to new healthInfo
+	ret := float64(0.0)
```
jzzheng22, line 5:
```suggestion
	ret := 0.0
```

jzzheng22 (changes requested):
Thanks! FYI, If you're using float64, you don't need to cast.

pkg/agents/team2/policies.go
laiwenq: Ahhh, I just clicked Replaced All.  Gonna remove them now.

laiwenq: Done! Have a look now. Thanks.

jzzheng22 (approved):
Thanks :)

hussain2603 (approved):
Thanks!

jzzheng22:
@tomeaton17  @laiwenq  feel free to merge btw :)

PR #151, opened 2021-12-29
Remove unnecessary mutex

jzzheng22:
# Summary
Don't need this mutex - we never call that function at all, but also we never modify the list of agents inside a goroutine.

### DIFF
--- a/pkg/infra/tower.go
+++ b/pkg/infra/tower.go
@@ -2,7 +2,6 @@ package infra
 
 import (
 	"math/rand"
-	"sync"
 
 	"github.com/SOMAS2021/SOMAS2021/pkg/messages"
 	"github.com/SOMAS2021/SOMAS2021/pkg/utils/globalTypes/day"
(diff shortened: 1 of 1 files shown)


pkg/infra/tower.go
```diff
@@ -121,8 +119,6 @@ func (t *Tower) ResetTower() {
 }
 
 func (t *Tower) TotalAgents() int {
```
JaafarRammal, line 121: This function is actually never used we could just remove it as well as `AgentsCount` in simEnv

jzzheng22: I don't have a problem with leaving it in - could be useful for logging later on. Happy to remove completely though.

JaafarRammal: (so can we make the PR "remove unused functions" and remove both `AgentCount` and `TotalAgents` all together?)

JaafarRammal: (sorry my GitHub doesn't show the new comment unless I refresh, feel free to do either)

JaafarRammal (approved):

PR #169, opened 2021-12-30
Error checking for config parameters, and sending errors over HTTP when in serve mode

### LINKED ISSUE #149: Error handling for config / filename
Empty body, division by 0, invalid filename

VFjr:
Closes #149 and #166

In the previous implementations, the code would try to run a simulation even if some vital parameters were not set. In some cases that could even cause division by 0. Now it checks that the vital parameters that have to be initialised are both initialised and not set to 0, and returns an error otherwise. 

When running in serve mode, if there is an error, it sends it as a 400 Bad Request. Here is a screenshot:

<details>
  <summary>Missing AgentHP in config, received a 400 Bad Request with the error message itself in the body </summary>

image

</details>


JaafarRammal (commented):
Legend. Please wait for a non-infra approval before merging

pkg/config/configparameters.go
```diff
@@ -68,7 +69,10 @@ func LoadParamFromJson(path string) (ConfigParameters, error) {
 		return tempParameters, err
 	}
 
-	CalculateDependantParameters(&tempParameters)
+	err = CalculateDependantParameters(&tempParameters)
```
jzzheng22, line 72:
```suggestion
	err = CalculateDependentParameters(&tempParameters)
```

pkg/config/configparameters.go
```diff
@@ -83,19 +87,99 @@ func LoadParamFromHTTPRequest(r *http.Request) (ConfigParameters, error) {
...
```
jzzheng22, line 90:
```suggestion
	err = CalculateDependentParameters(&tempParameters)
```

pkg/config/configparameters.go
```diff
@@ -83,19 +87,99 @@ func LoadParamFromHTTPRequest(r *http.Request) (ConfigParameters, error) {
...
+	if err != nil {
+		return tempParameters, err
+	}
 
 	return tempParameters, nil
 }
 
 //Some parameters depend directly on other parameters. This function calculates them and updates the original struct
-func CalculateDependantParameters(parameters *ConfigParameters) {
+func CalculateDependantParameters(parameters *ConfigParameters) error {
```
jzzheng22, line 99:
```suggestion
func CalculateDependentParameters(parameters *ConfigParameters) error {
```

jzzheng22 (changes requested):

jzzheng22 (approved):
ty bby

PR #172, opened 2021-12-30
Improve message logging

jzzheng22:
# Summary
Adds `String()` method so logging can print string type instead of an int.


hussain2603 (approved):
LGTM, thanks Jason :)

udai-arneja (dismissed):
LGTM

udai-arneja (dismissed):
🦾

pkg/messages/baseMessage_test.go
```diff
@@ -0,0 +1,96 @@
+package messages
+
```
woonmoon, line 2: Should we create a `test` package?

woonmoon (commented):
LGTM? Unsure if testing belongs in the messages package but otherwise LGTM

pkg/messages/baseMessage_test.go
JaafarRammal: I think we should be testing those components in a more connected way? like send a message and then receive it and then check it's been received correctly / to the right agent?

cc @jzzheng22

pkg/messages/baseMessage.go
```diff
@@ -23,6 +23,41 @@ const (
 	Response
 )
 
+func (m MessageType) String() string {
```
JaafarRammal, line 26: Can string be a function implemented in each message instead?

JaafarRammal (changes requested):
I think "String()" should be implemented under each message's implementation instead of a case in base. And every time a new message type is added, the String implementation is added too

pkg/messages/baseMessage_test.go
JaafarRammal: also @woonmoon I think in go that's how tests are done (you add a _test file in the same package), but I might be wrong

jzzheng22: Ideally we would test the message passing properly... but I'm just getting the easy one here for now. Can create an issue for longer testing... but also only two weeks left.

JaafarRammal:
> I think "String()" should be implemented under each message's implementation instead of a case in base. And every time a new message type is added, the String implementation is added too

On 2nd thoughts, that would mean we'd use `msg.String()` instead of `msg.MessageType().String()`, and the first is confusing since it could mean the message content. I believe this is the best way since:
- Enum key string needs the case
- Go doesn't allow function overload to have a String and INT version of the same function name for each implementation

JaafarRammal:
Pipeline goes brr

Makefile
```diff
@@ -17,15 +17,18 @@ test: dep ## Run unittests
 
 test-coverage: dep ## Run tests with coverage
 	@go test -coverprofile ./bin/cover.out -covermode=atomic ./...
-	@cat ./bin/cover.out
 
 build: dep ## Build the binary file
+	@go install golang.org/x/tools/cmd/stringer
```
JaafarRammal, line 22: Can these be removed now 👀

JaafarRammal (approved):

PR #180, opened 2021-12-31
Implemented Food Per Agent Ratio

### LINKED ISSUE #175: Scale maxPlatFood/FoodOnPlatform according to the number of agents in the tower.
In the spec it says that everyone should be able to survive (given the right strategies), but this isn't possible as the "FoodOnPlatform" is hard coded as 100 not proportional to number of agents

VFjr:
Closes #175 

Added `UseFoodPerAgentRatio` and `FoodPerAgentRatio` config parameters to allow to automatically set the amount of food depending on the number of agents. 

If `UseFoodPerAgentRatio` is true, then it multiplies `FoodPerAgentRatio` by number of agents to calculate total food. It ignores the `FoodOnPlatform` parameter if it is set

If  `UseFoodPerAgentRatio` is false, then it uses `FoodOnPlatform` and ignores `FoodPerAgentRatio`

I updated the config.json to include default parameters for them, and added error handling in case any of these three parameters are not set correctly.

### DIFF
--- a/pkg/config/configparameters.go
+++ b/pkg/config/configparameters.go
@@ -15,7 +15,7 @@ import (
 type ConfigParameters struct {
 	FoodOnPlatform food.FoodType `json:"FoodOnPlatform"`
 	Team1AgtOne    int           `json:"Team1AgtOne"`
-	Team1AgtTwo    int           `json:"Team2AgtTwo"`
+	Team1AgtTwo    int           `json:"Team1AgtTwo"`
 	Team2Agents    int           `json:"Team2Agents"`
 	Team3Agents    int           `json:"Team3Agents"`
 	Team4Agents    int           `json:"Team4Agents"`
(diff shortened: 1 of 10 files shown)


pkg/config/configparameters.go
```diff
@@ -13,35 +13,37 @@ import (
...
-	LogFileName    string        `json:"LogFileName"`
-	NumOfAgents    []int
-	NumberOfFloors int
-	TicksPerDay    int
-	DayInfo        *day.DayInfo
+	FoodOnPlatform       food.FoodType `json:"FoodOnPlatform"`
+	FoodPerAgentRatio    int           `json:"FoodPerAgentRatio"`
+	UseFoodPerAgentRatio bool          `json:"UseFoodPerAgentRatio"`
+	Team1AgtOne          int           `json:"Team1AgtOne"`
+	Team1AgtTwo          int           `json:"Team2AgtTwo"`
```
JaafarRammal, line 20: shouldn't that be `json:"Team1AgtTwo"`

JaafarRammal (dismissed):
Thanks. Would be great to have a wiki page for the config!

r15hil (dismissed):
LGTM thanks!

pkg/config/configparameters.go
VFjr: From a quick look this seems to be left from Team2's commit from PR #148. I saw that you also fixed it in your latest commit, and I incorporated those changes in my rebase.

VFjr:
> Thanks. Would be great to have a wiki page for the config!

Done, we now have a Config wiki page

jzzheng22 (approved):

PR #181, opened 2021-12-31
Active treaties memory

MattScottEEE:
# Summary

Implemented memory bank for agent treaties as well as helper functions

General flow:

When agents PROPOSE a treaty, this is ALWAYS pushed to the memory bank. It is when the response is received that this conditionally removes or increments the treaty

When agents REPLY to a treaty, we call a.RespondToTreaty (to prevent the import cycle of calling msg.Reply, since an agent pointer is needed) and conditionally push the treaty to memory.

## Additional Information

Added proposal ID to the newTreaty constructor

## Test Plan

Treaty generation and sending through Handle tested in team6


hussain2603 (commented):
I think it looks good. thanks!

pkg/infra/baseagent.go
```diff
@@ -213,6 +215,31 @@ func (a *Base) HealthInfo() *health.HealthInfo {
...
+
+func (a *Base) AddToActiveTreaties(treaty messages.Treaty) {
+	treatyID := treaty.Id()
+	a.activeTreaties[treatyID] = treaty
+}
+
+func (a *Base) DeleteFromActiveTreaties(treaty messages.Treaty) {
+	treatyID := treaty.Id()
+	delete(a.activeTreaties, treatyID)
+}
```
jzzheng22, line 230:
```suggestion
func (a *Base) DeleteTreaty(treatyID uuid.UUID) {
	delete(a.activeTreaties, treatyID)
}
```

pkg/infra/baseagent.go
```diff
@@ -222,5 +249,19 @@ func (a *Base) HandleResponse(msg messages.BoolResponseMessage)
...
 func (a *Base) HandleStateIntendedFoodTaken(msg messages.StateIntendedFoodIntakeMessage) {}
-func (a *Base) HandleProposeTreaty(msg messages.ProposeTreatyMessage)                    {}
-func (a *Base) HandleTreatyResponse(msg messages.TreatyResponseMessage)                  {}
+
+func (a *Base) HandleProposeTreaty(msg messages.ProposeTreatyMessage) {
+	treatyProposal := msg.Treaty()
+	// always add treaty to active list if proposing
+	a.AddToActiveTreaties(treatyProposal)
+}
+
```
woonmoon, line 258: I don't think this is correct, `HandleProposeTreaty` is called on the agent that receives a treaty proposal, not the agent who sent one. The handler needs to be over-writable by agents. I suggest leaving it blank and making sure later that agent teams call `AddToActiveTreaties` correctly.

pkg/infra/baseagent.go
```diff
@@ -222,5 +249,19 @@ func (a *Base) HandleResponse(msg messages.BoolResponseMessage)
...
 func (a *Base) HandleStateHP(msg messages.StateHPMessage)                                {}
 func (a *Base) HandleStateIntendedFoodTaken(msg messages.StateIntendedFoodIntakeMessage) {}
-func (a *Base) HandleProposeTreaty(msg messages.ProposeTreatyMessage)                    {}
-func (a *Base) HandleTreatyResponse(msg messages.TreatyResponseMessage)                  {}
+
+func (a *Base) HandleProposeTreaty(msg messages.ProposeTreatyMessage) {
+	treatyProposal := msg.Treaty()
+	// always add treaty to active list if proposing
+	a.AddToActiveTreaties(treatyProposal)
+}
```
jzzheng22, line 257:
```suggestion
func (a *Base) HandleProposeTreaty(msg messages.ProposeTreatyMessage) {}
```

pkg/infra/baseagent.go
```diff
@@ -222,5 +249,19 @@ func (a *Base) HandleResponse(msg messages.BoolResponseMessage)
...
+
+func (a *Base) HandleTreatyResponse(msg messages.TreatyResponseMessage) {
+	repliedTreaty := a.activeTreaties[msg.TreatyID()]
+	// conditionally remove treaty from active treaties once response is received
+	if msg.Response() {
+		repliedTreaty.SignTreaty()
+	} else {
+		a.DeleteFromActiveTreaties(repliedTreaty)
+	}
+}
```
jzzheng22, line 267:
```suggestion
func (a *Base) HandleTreatyResponse(msg messages.TreatyResponseMessage) {}
```

jzzheng22: Called when an agent receives a TreatyResponseMessage

pkg/infra/baseagent.go
jzzheng22: Called when an agent receives a ProposeTreatyMessage

jzzheng22 (changes requested):
Handler functions in base agent need to be empty - they need to be implemented by agent teams.

pkg/infra/baseagent.go
woonmoon, line 267: This is incorrect...? At least we shouldn't `DeleteFromActiveTreaties` if the message got rejected. We don't know if this Treaty has been also sent in the opposite direction and been agreed to.

woonmoon (commented):
The additional functions to `BaseAgent` look good to me, but I think the handler functions are incorrect and there is no forced removal of expired treaties right now?

pkg/infra/baseagent.go
MattScottEEE: Would rather pass treaty as argument, but otherwise fine with this

pkg/infra/baseagent.go
```diff
@@ -213,6 +215,31 @@ func (a *Base) HealthInfo() *health.HealthInfo {
...
 }
 
+func (a *Base) ActiveTreaties() map[uuid.UUID]messages.Treaty {
+	return a.activeTreaties
+}
+
+func (a *Base) AddToActiveTreaties(treaty messages.Treaty) {
+	treatyID := treaty.Id()
+	a.activeTreaties[treatyID] = treaty
+}
```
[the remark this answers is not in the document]
MattScottEEE: fine with this :)

pkg/infra/baseagent.go
MattScottEEE: Treaties are always added to the bank if proposed. If the response is negative, we remove the treaty from the active list?

pkg/infra/baseagent.go
MattScottEEE: The only issue is that this is something that needs to happen identically for all agents (treaties must be pushed to memory, what happens after is the agent team's call), so should really be a base function. Process treaties should be a function that's overwrite-able

pkg/infra/baseagent.go
MattScottEEE: I presume this should conditionally add the treaty then?

pkg/infra/baseagent.go
MattScottEEE: this is probably wrong then, not sure what correct behaviour should be

pkg/infra/baseagent.go
MattScottEEE: > This is incorrect...? At least we shouldn't `DeleteFromActiveTreaties` if the message got rejected. We don't know if this Treaty has been also sent in the opposite direction and been agreed to.

This is the same thing as asking someone to renew a treaty. If I ask you to sign a treaty and you say yes, then you send the same one back and I decline, then you effectively have a treaty lasting 1 tick that instantly gets cancelled

MattScottEEE: > This is incorrect...? At least we shouldn't `DeleteFromActiveTreaties` if the message got rejected. We don't know if this Treaty has been also sent in the opposite direction and been agreed to.

This is the same thing as asking someone to renew a treaty. If I ask you to sign a treaty and you say yes, then you send the same one back and I decline, then you effectively have a treaty lasting 1 tick that instantly gets cancelled

pkg/infra/baseagent.go
```diff
@@ -213,6 +215,29 @@ func (a *Base) HealthInfo() *health.HealthInfo {
...
+func (a *Base) RespondToTreaty(msg messages.ProposeTreatyMessage, senderFloor int, response bool) messages.TreatyResponseMessage {
+	treatyToReview := msg.Treaty()
+	if response {
+		treatyToReview.SignTreaty()
+		a.AddToActiveTreaties(treatyToReview)
+	}
+	reply := msg.Reply(a.ID(), senderFloor, response)
+
+	return reply
+}
```
jzzheng22, line 239: I'm not sure this function is necessary - this would be what you do to in `HandleProposeTreaty()` I think

pkg/infra/baseagent.go
jzzheng22: Inside this function you would decide whether you want to accept the proposed treaty and then send the appropriate response.

jzzheng22 (changes requested):

jzzheng22:
Superceded by #205.

PR #185, opened 2021-12-31
Added Timeout functionality

### LINKED ISSUE #142: Add timeout for simulation and a success / fail message in HTTP response

VFjr:
Closes #142

Added a parameter `SimTimeoutSeconds` to limit the amount of time used for simulation, and return an error if it is exceeded. 

If it is running in serve mode, if timeout is exceeded then it will return a HTTP 500 Internal Server Error with a timeout message, as seen below

<details>
  <summary> Timeout error in serve mode </summary>

image

</details>

If it is running in a single simulation mode, if timeout is exceeded it both prints that there was a timeout error and appends it to the log (unsure if I should handle it differently, let me know if there is a better way)

### DIFF
--- a/main.go
+++ b/main.go
@@ -38,14 +38,31 @@ func main() {
 			}
 
 			// logFileName returned to be used in dashboard
-			logfileName := runNewSimulation(parameters)
+			logFileName := ""
+
+			//channel and goroutine used for timeouts
+			c1 := make(chan string, 1)
+
+			go func() {
+				filenametemp := runNewSimulation(parameters)
+				c1 <- filenametemp
+			}()
+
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -38,14 +38,31 @@ func main() {
 			}
 
 			// logFileName returned to be used in dashboard
-			logfileName := runNewSimulation(parameters)
+			var logFileName string
```
jzzheng22, line 41:
```suggestion
			logFileName := ""
```

jzzheng22 (dismissed):
i think it work i hope

jzzheng22 (dismissed):

jzzheng22 (approved):

PR #188, opened 2021-12-31
Infra directory endpoint

### LINKED ISSUE #171: [APP] "/directory" endpoint
Send HTTP req to backend “/directory” and backend responds with a list of all available logs

VFjr:
Closes #171 

When the application runs in serve mode, it will now also listen to the /directory endpoint, and when it receives a request, it will return a list of all the existing folders inside the log folder:

<details>
  <summary> /directory request </summary>

image

</details>

If the folder does not exist, it returns a HTTP 500 Internal Server Error with an error message:

<details>
  <summary> /directory request with no log folder existing </summary>

image

</details>

Also an additional change was to return the error resulted by being unable to put the result struct into json format for the response in the /simulate endpoint from 400 Bad Request to 500 Internal Server Error.

### DIFF
--- a/main.go
+++ b/main.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"math/rand"
 	"net/http"
 	"os"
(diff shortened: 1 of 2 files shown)


jzzheng22 (approved):

main.go
```diff
@@ -61,12 +62,38 @@ func main() {
...
+		http.HandleFunc("/directory", func(w http.ResponseWriter, r *http.Request) {
+			//read directory
+			files, err := ioutil.ReadDir("./logs")
+			if err != nil {
+				http.Error(w, "Unable to open logs folder. There might not be any logs created yet", http.StatusInternalServerError)
+				return
+			}
+			for _, f := range files {
+				fmt.Println(f.Name())
+			}
```
JaafarRammal, line 83: Y'all skipped a print :) @jzzheng22 @VFjr

PR #190, opened 2021-12-31
Infra/propogate messages

woonmoon:
To talk to agents that are several floors away from you, the messages need to "propagate" in the correct direction.
When sending a message now please additionally specify the target floor e.g. for `AskFoodTakenMessage`s
`NewAskFoodTakenMessage(senderID uuid.UUID, senderFloor int, targetFloor int)`
Now in the `Visit` function, if the message's target floor is not equal to the agent's current floor, the agent's `HandlePropogate()` function will be called.

Of course, agents _should_ have the ability to lie, break the chain of communication or distort the message if they wanted to. However, in the "honest" iterations we assumed cooperation which meant:
- An agent does not tamper with the message as it relayed it forward
- An agent relays the message in the correct direction
- The agent does not store any information regarding a message that was not intended for him

`BaseAgent` has implemented an honest, cooperative `HandlePropogate()` function which politely sends a message in the correct direction. We ask that agent teams do not override this function in honest runs but you are welcome to override for other experiments.

(Team 6's (@MattScottEEE) message passing was slightly affected, I just added `a.Floor()+1` as a targetFloor field for one of your messages)

### DIFF
--- a/pkg/agents/team1/agent1/agent.go
+++ b/pkg/agents/team1/agent1/agent.go
@@ -33,35 +33,35 @@ func (a *CustomAgent1) Run() {
 	switch r {
 	case 0:
 		msg := messages.NewAskFoodTakenMessage(a.ID(), a.Floor(), a.Floor()-2)
-		a.SendMessage(-1, msg)
+		a.SendMessage(msg)
 		a.Log("I sent a message", infra.Fields{"message": "AskFoodTaken", "floor": a.Floor()})
 	case 1:
 		msg := messages.NewAskHPMessage(a.ID(), a.Floor(), a.Floor()+3)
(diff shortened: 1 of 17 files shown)


JaafarRammal (dismissed):

jzzheng22 (changes requested):
Can we get rid of `direction` inside of `BaseAgent.SendMessage()`? I think it's really confusing - and it should be calculable from the message itself anyway.

hussain2603 (approved):
looks good. Thanks gowoon :)

jzzheng22 (approved):
Have tested briefly, looks good, we'll test it more thoroughly with our agents later.

jzzheng22 (approved):

PR #195, opened 2022-01-01
[APP] Infra http frontend to backend

udai-arneja:
# Summary

Frontend HTTP to backend, UI to be improved.
Tested with logs see Logs below.

## Test Plan
Screenshot 2022-01-01 at 14 33 51
Screenshot 2022-01-01 at 14 33 36


jzzheng22:
😍

frontend/src/Components/NewRunConfig/AdvancedSettings/AdvancedSettings.tsx
```diff
@@ -0,0 +1,111 @@
...
+    id="testModal"
+    data-backdrop="false"
+    tabIndex={-1}
+    aria-labelledby="staticBackdropLabel"
+    aria-hidden="true"
+    >
+      <div className="modal-dialog">
+      <div className="modal-content">
+        <Card interactive={true} elevation={Elevation.TWO}>
+        <FormGroup helperText="In Days..." label="Simulation Length" labelFor="text-input" >
```
JaafarRammal, line 34: These can be changed to a for loop and importing the fields from a config file to make it cleaner

frontend/src/Components/NewRunConfig/NewRunState.tsx
```diff
@@ -0,0 +1,32 @@
...
+import InitConfigState from "./SimConfig";
+
+function request(configJSON: string){
+    const requestOptions = {
+        method: 'POST',
+        headers: { 'Content-Type': 'application/json' },
+        body: configJSON
+    };
+    var host = window.location.protocol + "//" + window.location.host;
+    fetch(`${host}/simulate`, requestOptions)
```
JaafarRammal, line 13: Can add a .catch and if backend responds with an error, we show it (although this can be a different PR)

frontend/src/Components/NewRunConfig/Settings/Settings.tsx
```diff
@@ -0,0 +1,81 @@
...
+      aria-hidden="true"
+    >
+      <div className="modal-dialog">
+        <div className="modal-content">
+          <div className="modal-header">
+            <h5 className="bp3-heading">New Run Configuration</h5>
+            <Button className="bp3-minimal close" icon="cross" text="" data-dismiss="modal" aria-label="Close" />
+          </div>
+          <div className="modal-body">
+          <FormGroup helperText="In Days..." label="Simulation Length" labelFor="text-input" labelInfo="(required)">
```
JaafarRammal, line 38: Same as before, can be changed into a for loop and load from config of inputs

JaafarRammal (commented):
Overall LGTM, amazing functionality added!
Some things can be improved for:
- Form: two or even 3 inputs per row would be better from smaller screens
- Form: inputs can be rendered in a loop to avoid duplication and it's better for maintaining
- POST: no error handling, but this can be added in a separate PR

frontend/src/Components/NewRunConfig/AdvancedSettings/AdvancedSettings.tsx
```diff
@@ -0,0 +1,75 @@
+import { Button, Checkbox, FormGroup, NumericInput } from "@blueprintjs/core";
+import { SubmitSimulation } from "../NewRunState";
+import { advancedParams, params } from "../ParameterLabels";
+import { SimConfig } from "../../../Helpers/SimConfig";
+import "./AdvancedSettings.css";
+
+export default function AdvancedSettingsMenu(state: any) {
```
JaafarRammal, line 7: You can pass an interface with the props definition to ensure the type of state[0] and state[1] is always the same and not dependent on ordering

frontend/src/Components/NewRunConfig/NewRunState.tsx
```diff
@@ -0,0 +1,38 @@
...
+    body: configJSON,
+  };
+  var host = window.location.protocol + "//" + window.location.host;
+  fetch(`${host}/simulate`, requestOptions)
+    .then(function (response) {
+      return response.json();
+    })
+    .catch(function (error) {
+      console.log("There has been a problem with your fetch operation: " + error.message);
+    });
```
JaafarRammal, line 19: The return handler doesn't really fit since you're returning a promise but not handling it anywhere. I suggest removing it completely
The catch can pop a toaster with "danger" intent to display the error message to the user

frontend/src/Components/Navbar.tsx
```diff
@@ -12,10 +12,18 @@ export default function Nav() {
...
           icon="search-around"
-          text="New Run"
+          text="Quick Run"
           data-toggle="modal"
           data-target="#exampleModal"
         />
+        <Navbar.Divider />
+        <Button
+          className="bp3-minimal"
+          icon="search-around"
```
JaafarRammal, line 22: I'm annoying but a different icon for each is better :p

frontend/src/Components/NewRunConfig/ParameterLabels.ts
```diff
@@ -0,0 +1,167 @@
...
+    {
+      helperText: "Base this on the size of your simulation.",
+      label: "TimeOut",
+      labelFor: "text-input",
+      labelInfo: "",
+      key: "SimTimeoutSeconds",
+    }
+];
+
+export interface Parameters {
```
JaafarRammal, line 161: "Parameter" instead of "Parameters"?

frontend/src/Components/NewRunConfig/Settings/Settings.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+export default function Settings(state: any) {
+  const config = state[0];
+  const setConfig = state[1];
+
+  function configHandler<Key extends keyof SimConfig>(value: any, keyString: any) {
+    var key: Key = keyString; // converting keyString to type Key
+    config[key] = value;
+    setConfig(config);
+    console.log(config);
```
JaafarRammal, line 15: Better to remove logs usually unless it's useful debugging in prod

frontend/src/Components/NewRunConfig/Settings/Settings.tsx
```diff
@@ -0,0 +1,73 @@
...
+                <NumericInput placeholder={config[i.key]} onValueChange={(value) => configHandler(value, i.key)} />
+              </FormGroup>
+            ))}
+            <FormGroup>
+              <Checkbox
+                label="Save Main"
+                type="checkbox"
+                onChange={(value) => configHandler((value.target as HTMLInputElement).checked, "LogMain")}
+              />
+            </FormGroup>
```
JaafarRammal, line 50: Now that I think about it, we never want to save that huge log file on our backend :)) so I think we can remove the option for users and just set to false by default (or we can disable it on the hosted version and they can set it to true locally only)

JaafarRammal (commented):
Added a few comments, will test now locally

JaafarRammal (approved):

JaafarRammal (approved):

PR #200, opened 2022-01-01
Update health parameters for more meaningful simulation results

mdubied:
# Summary

Update health parameters for more meaningful simulation results (in `config.json`)

## Additional Information

With these parameters, we need 2 food to get an HP increase of HPReqCtoW=5. 2 food is also the amount required to stay constant at the WeakLevel (if you are at HP=WeakLevel=10 and you eat 2 food, you have a final HP of 9 after applying the cost of living, but as 9>=HPCritical+HPReqCToW, you will wake up at HP=10 on the following day). 

Moreover, we go from HP=10 to HP=92 in 4 days if eating 2*tau, and need 5 days to reach HPMax=100. 

Finally, it takes 8 days to go from maxHP=100 to HPCritical when eating nothing, and therefore 11 days to die if we keep MaxDaysCritical=3.

## Test Plan

A Matlab code was run to determine better parameters than the current ones.


JaafarRammal (commented):
I would suggest to set these to defaults in the config handler instead of config.json, so if the values are not passed,  they are loaded by default, although happy to just have them in config.json for now

JaafarRammal (approved):
Re: handling defaults should be done by us in infra, ill approve

udai-arneja:
@JaafarRammal  I think having them in the config file is alright tbh, either, we'll allow changes from the user or (time willing) we can have different modes: e.g. hard mode =  we need 4 food to get an HP increase of HPReqCtoW=5 and easy mode =  we need 1 food to get an HP increase of HPReqCtoW=5, medium most = as above.

udai-arneja (approved):

PR #203, opened 2022-01-02
[APP] Frontend logs

JaafarRammal:
# Summary

Code that allows to easily add logs to frontend. Can directly load logs now from the logs folder and show them as wished.

image

**Contributing**

Backend:
- First, add a logger in `statelog.go` with a logging function
- In the backend code, call your logging function accordingly when you want to log something (e.g. if you are making a death logger, you would call your death logging function when an agent dies and log something like the agent type or the agent age when it died)
- The logger will automatically save this data to a separate file

Example:

Extend `statelog.go` with your new logger and any variables to keep log state (here death count)

```golang
type StateLog struct {
//...
	deathLogger *log.Logger
        deathCount int
//...
}

func NewLogState(folderpath string, saveMainLog bool) *StateLog {
//...
	deathLogger, err := l.AddLogger("death", "death.json")
	handleNewLoggerErr(err)
//...
	return &StateLog{
//...
		deathLogger: deathLogger,
//...
	}
}

func (ls *StateLog) LogAgentDeath(day int, tick int, agentType agent.AgentType) {
	ls.deathCount++
	ls.deathLogger.WithFields(log.Fields{"day": day, "tick": tick, "agent_type": agentType.String(), "cumulativeDeaths": ls.deathCount}).Info("")
}
```

Now you can call `sE.stateLog.LogAgentDeath` or `t.stateLog.LogAgentDeath` (they're pointing to the same logger)

Frontend:
- In the frontend, create a log type that extends `Log.ts`
- Create a fetching function that parses the backend logged file into this log type
- Add your new type to the Result interface

Example:

Create a new file in `frontend/src/Helpers/Logging` which has your interface and parser:

```typescript
import { GetFile } from "../API";
import { Log } from "./Logging";

export interface DeathLog extends Log{
  cumulativeDeaths: number
  agentType: string
}

export function GetDeathLogs(filename: string): Promise<DeathLog[]> {
  return new Promise<DeathLog[]>((resolve, reject) => {
    GetFile(filename, "death")
      .then((deaths) =>
        resolve(
          deaths.map(function (e: any) {
            const d: DeathLog = {
              // this is the parsing into our new DeathLog type
              cumulativeDeaths: e["cumulativeDeaths"],
              agentType: e["agent_type"],
              tick: e["tick"],
              day: e["day"],
            };
            return d;
          })
        )
      )
      .catch((err) => reject(err));
  });
}
```

Extend the Result interface

``` typescript
// Result.ts

import { DeathLog } from "./Logging/Death";
import { FoodLog } from "./Logging/Food";

export interface Result {
  title: string
  deaths: DeathLog[] // adding to Results the new log
  food: FoodLog[]
}
```

Use the result by extending the GetResult API function

```typescript
export function GetResult(filename: string): Promise<Result> {
  return new Promise<Result>(async (resolve, reject) => {
//...
    // deaths
    var deaths: DeathLog[] = [];
    promises.push(GetDeathLogs(filename).then((d) => (deaths = d)));
//...
    Promise.all(promises).then((_) =>
      resolve({
//...
        deaths: deaths,
//...
      })
    );
  });
}
```


frontend/src/Components/Results.tsx
```diff
@@ -1,66 +1,67 @@
...
-          <i>Select an exsiting simulation result to view results</i>
-        </H6>
-      )}
+      {loading && <Spinner intent="primary" />}
+      {!loading &&
+        (logName !== "" && result ? (
+          <ResultDisplay result={result} />
+        ) : (
+          <H6>
+            <i>Select an exsiting simulation result to view results</i>
```
jzzheng22, line 36:
```suggestion
            <i>Select an existing simulation result to view results</i>
```

jzzheng22 (dismissed):
Note: Not approving any of the TypeScript code - please wait for a frontend review for that.

JaafarRammal:
Updated some errors concerning empty log files / empty arrays. Thanks @jzzheng22 for testing 😅

udai-arneja (approved):
less go

PR #204, opened 2022-01-02
Fix FoodRequired function in health

jzzheng22:
# Summary
The `health.FoodRequired` function was not taking into account an agent's daily hp decay. This has now been fixed.

@MattScottEEE  I've copied over the old version and put it into your agent so you're using the original implementation - not sure what you want to do with your agent.


pkg/utils/globalTypes/health/health.go
```diff
@@ -38,10 +39,24 @@ func NewHealthInfo(parameters *config.ConfigParameters) *HealthInfo {
 		MaxDayCritical: parameters.MaxDayCritical,
 		HPLossBase:     parameters.HPLossBase,
 		HPLossSlope:    parameters.HPLossSlope,
+		maxPlatFood:    parameters.FoodOnPlatform,
 	}
 }
 
+// This function takes in an initial HP and a goal HP, and returns how much food you should eat to
+// achieve the goal HP, while taking into account HP decay.
+// This function should be called only for HP values above WeakLevel.
```
JaafarRammal, line 48: > This function should be called only for HP values above WeakLevel.
What's the behavior if it's called on that? Undefined or breaks? Could add an error to the function (if it's not too hard of course)

JaafarRammal (commented):
Code structure LGTM, one minor comment only for the function behavior

pkg/utils/globalTypes/health/health.go
jzzheng22: Gonna say it's undefined - because for critical HP the behaviour of an agent's HP is defined on the number of days they spend in critical

pkg/utils/globalTypes/health/health.go
```diff
@@ -38,10 +39,24 @@ func NewHealthInfo(parameters *config.ConfigParameters) *HealthInfo {
...
 func FoodRequired(currentHP int, goalHP int, healthInfo *HealthInfo) food.FoodType {
-	denom := healthInfo.Width - float64(goalHP) + (1-healthInfo.HPLossSlope)*float64(currentHP) - float64(healthInfo.HPLossBase) + healthInfo.HPLossSlope*float64(healthInfo.WeakLevel)
-	return food.FoodType(healthInfo.Tau * math.Log(healthInfo.Width/denom))
+	slope := healthInfo.HPLossSlope
+	base := healthInfo.HPLossBase
+	numer := float64(goalHP) - (1-slope)*float64(currentHP) + float64(base) - slope*float64(healthInfo.WeakLevel)
+	denom := healthInfo.Width * (1 - slope)
+	food := food.FoodType(-1 * healthInfo.Tau * math.Log(1-numer/denom))
```
hussain2603, line 54: maybe I did my maths wrong, but according to your equations when I put current HP = 15 and goal HP = 100, I get numer (91) > denom (38.4) and so the log would be undefined?

 im not sure how go handles undefined values and maybe that's why you have the if statement below this line?

hussain2603 (changes requested):

pkg/utils/globalTypes/health/health.go
hussain2603: I mean I can see the if statement below the line so I guess that somehow handles the undefined log. and I see the test so I guess it works but not completely sure.

hussain2603 (approved):
not too sure about the undefined log, but the test cases work so I guess it works.

PR #209, opened 2022-01-02
Cleaned up console prints

### LINKED ISSUE #189: Clean up console prints for backend
Well, I accidentally left some prints in my last PR that are not needed, so I'll fix that tomorrow. 

Also there is currently a mix of fmt.Println and log.Println throughout the code. Jaafar suggested to switch everything to log.Println, and also add some new useful prints, such as saying when a simulation started, to what file it was saved, how long did it take, etc

VFjr:
Closes #189. 

Removed some unnecessary prints related to directories. They were left from a template i used. Oops.

Also cleaned up the console prints by using log.Info and log.Error instead of fmt.Println 

Edit: Also renamed logFileName to LogFolderName everywhere except in the json parsing, so it is still compatible with the current frontend, and also now the directory array is initialised to empty instead of nil (as per Jaafar's request)

### DIFF
--- a/main.go
+++ b/main.go
@@ -44,27 +44,33 @@ func main() {
 			parameters, err := config.LoadParamFromHTTPRequest(r)
 			if err != nil {
 				http.Error(w, err.Error(), http.StatusBadRequest)
+				log.Error("Unable to load parameters from the Simulation Request: " + err.Error())
 				return
 			}
 
-			// logFileName returned to be used in dashboard
-			logFileName := ""
+			rand.Seed(time.Now().UnixNano())
+			logFolderName, err := setupLogFile(parameters.LogFolderName, parameters.LogMain)
+			if err != nil {
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				log.Error("Unable to setup log file: " + err.Error())
+			}
 
 			//channel and goroutine used for timeouts
 			c1 := make(chan string, 1)
 
 			go func() {
-				filenametemp := runNewSimulation(parameters)
-				c1 <- filenametemp
+				log.Info("Simulation " + logFolderName + " started")
+				runNewSimulation(parameters, logFolderName)
+				c1 <- "Simulation Finished"
 			}()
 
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -168,9 +164,9 @@ func main() {
 		// Listen on our channel AND a timeout channel - which ever happens first.
 		select {
 		case <-c1:
-			fmt.Println("Simulation Finished Successfully")
+			log.Info("Simulation Finished Successfully")
 		case <-time.After(time.Duration(parameters.SimTimeoutSeconds) * time.Second):
-			fmt.Println("Simulation Timeout")
+			log.Info("Simulation Timeout")
```
jzzheng22, line 169: If you're logging in fatal why are you logging in info?

jzzheng22 (changes requested):

main.go
```diff
@@ -168,9 +164,9 @@ func main() {
 		// Listen on our channel AND a timeout channel - which ever happens first.
 		select {
 		case <-c1:
-			fmt.Println("Simulation Finished Successfully")
+			log.Info("Simulation Finished Successfully")
```
JaafarRammal, line 167: Can we add the simulation name in the message and a "Simulation {NAME} started"?

JaafarRammal (commented):
Would it be possible to add logs as well in the HTTP handlers? Would be useful!

main.go
VFjr: No idea where that came from lol. Thanks

main.go
VFjr: Wouldn't it be reasonable to expect the simulation name to be the same as the folder name that is passed already?

JaafarRammal: Oh true this is single local simulation, my bad!
It would be useful to log folder name for the http handler

jzzheng22 (dismissed):

jzzheng22:
just a comment, it's probably also fine to use `log.Fatal()` here since if there are errors at the start you don't want to run the simulation right?

jzzheng22 (approved):

PR #222, opened 2022-01-02
Timeouts now end the goroutines rather than keeping them running until the end

VFjr:
So up until now, when a timeout happened, main would show that there was a timeout, but the goroutine would actually keep running in the background until it was complete, wasting resources.

It turns out there is no built in way to kill a thread / goroutine in golang. So i had to do it manually, by checking if timeout was triggered every tick. Might not be the neatest implementation, but it works and doesn't seem too bad. If you have any recommendations let me know.

Tested using manual prints to ensure that once a timeout was triggered, the simulate and simulationLoop functions would not continue.

### DIFF
--- a/pkg/simulation/simIteration.go
+++ b/pkg/simulation/simIteration.go
@@ -1,6 +1,7 @@
 package simulation
 
 import (
+	"context"
 	"sync"
 
 	"github.com/SOMAS2021/SOMAS2021/pkg/infra"
@@ -18,7 +19,7 @@ func (sE *SimEnv) World() world.World {
 	return sE.world
 }
 
-func (sE *SimEnv) simulationLoop(t *infra.Tower) {
+func (sE *SimEnv) simulationLoop(t *infra.Tower, ctx context.Context, ch chan<- string) {
 	t.Reshuffle()
 	for sE.dayInfo.CurrTick <= sE.dayInfo.TotalTicks {
 		sE.Log("", Fields{"Current Simulation Tick": sE.dayInfo.CurrTick})
@@ -28,7 +29,16 @@ func (sE *SimEnv) simulationLoop(t *infra.Tower) {
 		sE.replaceAgents(t)
 		// t.TowerStateLog(" end of tick")
 		sE.dayInfo.CurrTick++
+
+		//returns if there is a timeout
+		//continously checks every tick since there is no way to kill a goroutine
+		select {
+		case <-ctx.Done():
+			return
+		default:
+		}
 	}
+	ch <- "Simulation Finished"
 }
 
 func (sE *SimEnv) TowerTick() {
--- a/pkg/simulation/simulation.go
+++ b/pkg/simulation/simulation.go
@@ -1,6 +1,8 @@
 package simulation
 
 import (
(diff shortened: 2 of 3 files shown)


pkg/simulation/simIteration.go
```diff
@@ -28,7 +29,17 @@ func (sE *SimEnv) simulationLoop(t *infra.Tower) {
...
 		// t.TowerStateLog(" end of tick")
 		sE.dayInfo.CurrTick++
+
+		//returns if there is a timeout
+		//continously checks every tick since there is no way to kill a goroutine
+		select {
+		default: //nothing
+		case <-ctx.Done():
+			return
+		}
```
jzzheng22, line 39: Is there a reason the default is before the case?

pkg/simulation/simulation.go
```diff
@@ -73,7 +75,15 @@ func (sE *SimEnv) Simulate() {
...
 	sE.Log("Simulation Started")
-	sE.simulationLoop(t)
+	sE.simulationLoop(t, ctx, ch)
+
+	//returns if there was a timeout
+	select {
+	default: //nothing
+	case <-ctx.Done():
+		return
+	}
```
jzzheng22, line 85: Is there a reason the default is before the case?

pkg/simulation/simIteration.go
```diff
@@ -28,7 +29,17 @@ func (sE *SimEnv) simulationLoop(t *infra.Tower) {
...
 	}
+	//only gets here if there was no timeout
```
jzzheng22, line 41:
```suggestion
```

pkg/simulation/simIteration.go
```diff
@@ -28,7 +29,17 @@ func (sE *SimEnv) simulationLoop(t *infra.Tower) {
 		sE.replaceAgents(t)
 		// t.TowerStateLog(" end of tick")
 		sE.dayInfo.CurrTick++
+
+		//returns if there is a timeout
+		//continously checks every tick since there is no way to kill a goroutine
+		select {
+		default: //nothing
```
jzzheng22, line 36:
```suggestion
		default:
```

pkg/simulation/simulation.go
```diff
@@ -73,7 +75,15 @@ func (sE *SimEnv) Simulate() {
 	sE.generateInitialAgents(t)
 
 	sE.Log("Simulation Started")
-	sE.simulationLoop(t)
+	sE.simulationLoop(t, ctx, ch)
+
+	//returns if there was a timeout
+	select {
+	default: //nothing
```
jzzheng22, line 82:
```suggestion
	default:
```

jzzheng22 (changes requested):

pkg/simulation/simIteration.go
VFjr: Well, I am not sure. It was like this in some template i followed. Will check if it works if i switch them around

VFjr: Yeah, it still works fine with them the usual way around

jzzheng22 (approved):

PR #224, opened 2022-01-03
Override HandlePropogate message function to process State... messages

benstobbs:
# Summary
Overrides the base agent's HandlePropogate message function to manually run our Handle... functions on any State... messages, resulting in our agents "remembering" information which we pass on.


Rhys-J (approved):
LGTM, thanks Ben

PR #230, opened 2022-01-04
[Team 5] Implement treaty handling

Rhys-J:
# Summary

Added handler for proposed treaties, added in decisions of when to reject and accept
Added function to check if we are in treaty and alter the amount of taken food if necessary
Added functions to check for conflicting treaties
Added proposal of a treaty if certain criteria is met

## Additional Information

Changes to ending death log to print dead agents included

## Test Plan

Simulations ran with our own agents and with other agent teams


pkg/agents/team5/messages.go
```diff
@@ -206,3 +206,91 @@ func (a *CustomAgent5) HandleStateIntendedFoodTaken(msg messages.StateIntendedFo
...
+	treaty := msg.Treaty()
+	if len(a.ActiveTreaties()) > 0 || treaty.Request() == messages.Inform {
+		//TODO: Have a checkForConflicts function that determines if we can feasibly sign up to multiple treaties.
+		//For now, only agree to sign up to 1 treaty at a time for simplicity
+		//Also, reject any inform treaties as I'm not sure what they are or how to handle them
+		a.RejectTreaty(msg)
+		return
+	}
+
+	if (treaty.Request() == messages.LeaveAmountFood && treaty.RequestOp() == messages.EQ) || treaty.RequestOp() == messages.LE || treaty.RequestOp() == messages.LT {
```
benstobbs, line 226: might be wrong, but aren't the brackets in the wrong place here? shouldn't it be:
`if treaty.Request() == messages.LeaveAmountFood && (treaty.RequestOp() == messages.EQ || treaty.RequestOp() == messages.LE || treaty.RequestOp() == messages.LT) {`
or (slightly) more simply:
`if treaty.Request() == messages.LeaveAmountFood && treaty.RequestOp() != messages.GT && treaty.RequestOp != messages.GE {`

pkg/agents/team5/messages.go
```diff
@@ -206,3 +206,91 @@ func (a *CustomAgent5) HandleStateIntendedFoodTaken(msg messages.StateIntendedFo
...
+	if (treaty.ConditionOp() == messages.LE || treaty.ConditionOp() == messages.LT) && (treaty.Condition() == messages.HP || treaty.Condition() == messages.AvailableFood) {
+		//Reject treaties that have a condition of being less than something
+		// Why agree to treaty which is bounded by you having lower HP
+		// Why agree to treaty which is bounded by the amount of food on platform being low
+		// All of these are conditions where you will be desperate for food so shouldn't be limitting yourself if you do get the opurrtunity to eat
+		a.RejectTreaty(msg)
+		return
+	}
+
+	if (treaty.ConditionOp() == messages.GE || treaty.ConditionOp() == messages.GT) && treaty.Condition() == messages.Floor {
```
benstobbs, line 242: with GE and GT, doesn't this currently reject treaties that are bounded by you being higher in the tower? I think this might be the wrong way around

pkg/agents/team5/messages.go
Rhys-J: My thinking was this: 
Code is meant to reject all treaties with less than (LE or LT) requests, but only reject equal to (EQ) requests if the request type is for a specific food amount. We would still accept treaties for equal to (EQ) requests if the request type is for a percentage amount of food.

pkg/agents/team5/messages.go
```diff
@@ -206,3 +206,91 @@ func (a *CustomAgent5) HandleStateIntendedFoodTaken(msg messages.StateIntendedFo
...
+	}
+
+	if treaty.Condition() == messages.HP && treaty.ConditionValue() < a.HealthInfo().WeakLevel*2 {
+		//Reject any HP condition based treaty if the condition is too strict, and you would be put into critical by not eating
+		a.RejectTreaty(msg)
+		return
+	}
+
+	if treaty.Condition() == messages.Floor && treaty.ConditionValue() != 1 && treaty.Duration() >= a.HealthInfo().MaxDayCritical {
+		//Reject any floor condition that involves us not being on the top floor and lasts for more than days you can survive in critical
```
benstobbs, line 255: This seems a bit too broad? This would reject things like the first few floors leaving high amounts of food, which would be reasonable treaties.

pkg/agents/team5/messages.go
Rhys-J: I wasn't sure how to interpret the floor values in terms of GT or LT. As being higher up in the tower means having a floor value less than (LT) those below you. So if the condition is GE or GT than it means being lower down in the tower as the floor value you have will be larger. Not sure if that makes sense?

pkg/agents/team5/messages.go
benstobbs: yeah ok, definitely agree it's good for now

pkg/agents/team5/messages.go
benstobbs: ah ok, yes I guess the only operations are to leave food so this makes sense!

jzzheng22:
Note for later: Change call to `resetSocialKnowledge` to occur only on Day6.

pkg/agents/team5/agent.go
```diff
@@ -100,6 +100,63 @@ func (a *CustomAgent5) checkForLeader() {
...
+
+	if t1.Condition() == t2.Condition() && !conditionsIntersect {
+		// If the treaties are based on the same type of condition, they can coexist
+		// when the conditions are mutually exclusive (can never occurr simultaneously)
+		return true
+	}
+
+	// Otherwise, if the treaties are based on different conditions, or the conditions
+	// can occurr simultaneously, they can coexist only if the requests can be fulfilled simultaneously
+	return a.statementsIntersect(t1.RequestOp(), t1.RequestValue(), t2.RequestOp(), t2.RequestValue())
```
benstobbs, line 148: this assumes the Request()s themselves are the same, which is incorrect.

pkg/agents/team5/agent.go
```diff
@@ -86,6 +93,114 @@ func (a *CustomAgent5) checkForLeader() {
...
+func (a *CustomAgent5) statementsIntersect(op1 messages.Op, value1 int, op2 messages.Op, value2 int) bool {
+	switch op1 {
+	case messages.EQ:
+		return ((op2 == messages.EQ && value1 == value2) ||
+			(op2 == messages.LT && value1 < value2) ||
+			(op2 == messages.LE && value1 <= value2) ||
+			(op2 == messages.GT && value1 > value2) ||
+			(op2 == messages.GE && value1 >= value2))
+	case messages.LT:
+		return ((op2 == messages.EQ && value1 < value2) ||
```
Rhys-J, line 112: Please correct me if I'm getting this wrong. If the first condition is to be less than value1, and the second is to be equal to value 2, they would intersect if value1 > value 2 wouldn't they???

pkg/agents/team5/agent.go
```diff
@@ -86,6 +93,114 @@ func (a *CustomAgent5) checkForLeader() {
...
+			op2 == messages.LT ||
+			op2 == messages.LE ||
+			(op2 == messages.GT && (value1-1) > value2) || // need the -1 here to disallow x<3 AND x>2
```
Rhys-J, line 115: If every condition value is an integer then we shouldn't need the -1 here. x<3 would be 0, 1, 2. x>2 would be 3, 4 ... Which don't intersect

Rhys-J: I think you want to use value1-2 > value2. If value1 is 4, and value2 is 2. These interesect at 3.

pkg/agents/team5/agent.go
```diff
@@ -86,6 +93,114 @@ func (a *CustomAgent5) checkForLeader() {
...
+			(op2 == messages.GE && value1 > value2))
```
Rhys-J, line 116: I think this needs to have the -1. If value1 = 3, they would intersect if value2 is 2 or less.
value1-1 > value2

pkg/agents/team5/agent.go
```diff
@@ -86,6 +93,114 @@ func (a *CustomAgent5) checkForLeader() {
...
+	case messages.LE:
+		return (op2 == messages.LT ||
+			op2 == messages.LE ||
+			value1 > value2)
```
Rhys-J, line 120: More checks needed here
op2 == EQ && value1 >= value2
op2 == GT && value1 > value2
op2 == GE && value1 >= value2

These should all return true

pkg/agents/team5/agent.go
```diff
@@ -86,6 +93,114 @@ func (a *CustomAgent5) checkForLeader() {
...
+	case messages.GT:
+		return ((op2 == messages.EQ && value1 > value2) ||
```
Rhys-J, line 122: Same as comment above, I think these would intersect if value1 < value2

Rhys-J:
Note: Consider duration of proposed treaty in decision to accept, shorter treaties should be more likely to be accepted

pkg/agents/team5/agent.go
benstobbs: yes these don't intersect, but would be marked as intersecting if the -1 was not there because for `x<3 AND x>2`,
`value1 = 3`, `value2 = 2`, returning `3 > 2 = true`
therefore I'm pretty sure we need the -1 to make `2 > 2 = false`
this is getting very confusing though aha

pkg/agents/team5/agent.go
benstobbs: same logic as above, agree that -1 is required

pkg/agents/team5/agent.go
benstobbs: yes, I will flip these!

pkg/agents/team5/agent.go
benstobbs: will fix by rejecting treaties of the same request type at this point

pkg/agents/team5/agent.go
benstobbs: agree, will fix

pkg/agents/team5/agent.go
benstobbs: yep, not sure how I missed that, thanks!

pkg/agents/team5/messages.go
```diff
@@ -13,6 +13,31 @@ func (a *CustomAgent5) getMessages() {
...
+		sendToFloor = a.Floor() + 2
+	case 4:
+		sendToFloor = a.Floor() - 2
+	case 5:
+		sendToFloor = a.Floor() + 3
+	case 6:
+		sendToFloor = a.Floor() - 3
+	}
+
+	msg := messages.NewProposalMessage(a.ID(), a.Floor(), sendToFloor, *a.currentProposal)
```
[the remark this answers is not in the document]
Rhys-J: There is not currently, if an agent tries to send a message to floor < 1 or > number of floors the message send will just fail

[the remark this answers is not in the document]
daryllimyt: I see, no worries then

pkg/agents/team5/agent.go
```diff
@@ -74,11 +88,137 @@ func (a *CustomAgent5) updateSatisfaction() {
...
+			a.overrideCalculation(treaty)
+		case treaty.Condition() == messages.Floor && treaty.ConditionOp() == messages.LE && a.Floor() <= treaty.ConditionValue():
+			a.overrideCalculation(treaty)
+		case treaty.Condition() == messages.AvailableFood && treaty.ConditionOp() == messages.EQ && a.CurrPlatFood() == food.FoodType(treaty.ConditionValue()):
+			a.overrideCalculation(treaty)
+		case treaty.Condition() == messages.AvailableFood && treaty.ConditionOp() == messages.GT && a.CurrPlatFood() > food.FoodType(treaty.ConditionValue()):
+			a.overrideCalculation(treaty)
+		case treaty.Condition() == messages.AvailableFood && treaty.ConditionOp() == messages.GE && a.CurrPlatFood() >= food.FoodType(treaty.ConditionValue()):
+			a.overrideCalculation(treaty)
+		}
```
jzzheng22, line 212: TODO: add a fallthrough statement here

pkg/agents/team5/messages.go
```diff
@@ -13,149 +14,314 @@ func (a *CustomAgent5) getMessages() {
...
+			// Reject treaty in which you would not be able to eat enough food to avoid critical level
+			a.RejectTreaty(msg)
+			return
+		}
+		if treaty.Request() == messages.LeavePercentFood && treaty.ConditionValue()*int(float64(100-treaty.RequestValue())/100) < 3 {
+			// Reject treaty in which you would not be able to eat enough food to avoid critical level
+			a.RejectTreaty(msg)
+			return
+		}
+	}
```
jzzheng22, line 234: TODO: change to switch statement and use fallthrough

pkg/agents/team5/agent.go
Rhys-J: Please see latest commit to see if resolved

SaraFFdez (dismissed):
LGTM

pkg/agents/team5/treaty.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+}
+
+func (a *CustomAgent5) HandleTreatyResponse(msg messages.TreatyResponseMessage) {
+	if msg.Response() {
+		treaty := a.ActiveTreaties()[msg.TreatyID()]
+		a.ActiveTreaties()[msg.TreatyID()] = treaty
+		a.addToSocialFavour(msg.SenderID(), a.socialMemory[msg.SenderID()].favour+2)
+	}
+}
```
JaafarRammal, line 160: I'm missing something or an agent could break your agent by sending a response message but modify the treaty id to one you don't have? xD `treaty := a.ActiveTreaties()[msg.TreatyID()]`

Just checking if the memory accesses in this function are always valid

pkg/agents/team5/social.go
```diff
@@ -77,11 +78,20 @@ func (a *CustomAgent5) calculateAverageFavour() int {
...
+// TODO: Consider adding check for exists and if not exists then add to memory.
+func (a *CustomAgent5) resetDaysSinceLastSeen(id uuid.UUID) {
+	mem := a.socialMemory[id]
+	mem.daysSinceLastSeen = 0
+	a.socialMemory[id] = mem
+}
+
 // TODO: Consider adding check for exists and if not exists then add to memory.
 func (a *CustomAgent5) resetSocialKnowledge(id uuid.UUID) {
 	mem := a.socialMemory[id]
```
JaafarRammal, line 97: What's the behavior if not existing in memory?

JaafarRammal (approved):

pkg/agents/team5/treaty.go
Rhys-J: Please see 'Added check for active treaty in response handler' commit for fix

pkg/agents/team5/social.go
Rhys-J: Please see 'Added safety checks for memory existing in social functions' commit for fix

hussain2603 (approved):

pkg/agents/team5/utils.go
```diff
@@ -21,6 +29,115 @@ func (a *CustomAgent5) memoryIdExists(id uuid.UUID) bool {
...
+			log.Error("Simulation - team5/agent.go: \t Reached unreachable code in statementsIntersect", op1, value1, op2, value2)
+			return true
+		}
+
+	case messages.GT:
+		switch op2 {
+		case messages.EQ, messages.LE:
+			return value1 < value2
+		case messages.LT:
+			return value1 < value2-1
```
woonmoon, line 95: Change to `<=`?

pkg/agents/team5/utils.go
```diff
@@ -21,6 +29,115 @@ func (a *CustomAgent5) memoryIdExists(id uuid.UUID) bool {
...
+		}
+
+	case messages.LT:
+		switch op2 {
+		case messages.EQ, messages.GE:
+			return value1 > value2
+		case messages.LT, messages.LE:
+			return true
+		case messages.GT:
+			return value1-1 > value2
```
woonmoon, line 71: Change to `>=`

woonmoon (approved):

PR #233, opened 2022-01-04
[APP] Frontend refactor

JaafarRammal:
# Summary

Move sim call to API and refactor a few props


udai-arneja (approved):
Refactor looks good thanks Jaafar x

PR #242, opened 2022-01-05
Add min, max, and average age upon death statistics to deathlogger and frontend

daryllimyt:
Self explanatory, extended existing death logger to include this. Willing to take on additional statistics additions if anyone has anything they'd like to include.


jzzheng22:
What are you using to write code? Are you using VSCode?

daryllimyt:
yeah using vscode, idk why i keep having linting issues

jzzheng22:
> yeah using vscode, idk why i keep having linting issues

Do you have the Go extension installed properly? `gofmt` should run automatically when you save files.

daryllimyt:
yeah I think I had a different linter as my default, changed it and works now

frontend/src/Components/Results/Results.tsx
```diff
@@ -76,6 +76,26 @@ function ResultDisplay(props: ResultDisplayProps) {
...
+              description="Max agent age upon death"
+              title={Max(result.deaths.map((d) => d.ageUponDeath)).toString()}
+            />
+          </div>
+          <div className="col-lg-6">
+            <ReportCard
+              description="Min agent age upon death"
+              title={Min(result.deaths.map((d) => d.ageUponDeath)).toString()}
+            />
+          </div>
```
JaafarRammal, line 98: Since this is getting larger could we move it to a separate component `Stats.tsx` which takes in results in the props? Not necessary feel free to ignore and resolve the comment

JaafarRammal (approved):
Very straightforward thanks Daryl!

frontend/src/Components/Results/Results.tsx
daryllimyt: Yeah for sure can do

frontend/src/Helpers/Utils.ts
```diff
@@ -1,3 +1,9 @@
 export function Average(arr: number[]): number {
   return arr.reduce((a, b) => a + b) / arr.length;
 }
+export function Max(arr: number[]): number {
+  return arr.reduce((a, b) => Math.max(a, b));
+}
+export function Min(arr: number[]): number {
+  return arr.reduce((a, b) => Math.min(a, b));
+}
```
JaafarRammal, line 9: I might be wrong but you can just do
```ts
Math.max(...arr)
Math.min(...arr)
```

So I guess this utils are not needed and can just use the native `Math` directly?

JaafarRammal (changes requested):
One last comment about using the native `Math.max(...arr)` on arrays :) and then gtg

frontend/src/Helpers/Utils.ts
daryllimyt: you are probably right oops, i'll get that changed

JaafarRammal (approved):

PR #243, opened 2022-01-05
Round() before int casting

mdubied:
# Summary

Perform math.Round() before to cast floats to integers in the health functions (updateHP, HPDecay, FoodRequired). This operation allows for better performance of the FoodRequired function

## Additional Information

The precision gained by this method was analysed by using the best set of health parameters we have so far, in a matlab script:
tau=15;
w=48;
b=5;
s=0.2;
WeakLevel=10;
HPCritical=3;
HPReqCToW=5;

### DIFF
--- a/pkg/infra/baseAgent.go
+++ b/pkg/infra/baseAgent.go
@@ -153,7 +153,7 @@ func (a *Base) setHP(newHP int) {
 
 // Modeled as a first order system step answer (see documentation for more information)
 func (a *Base) updateHP(foodTaken food.FoodType) {
-	hpChange := int(a.tower.healthInfo.Width * (1 - math.Pow(math.E, -float64(foodTaken)/a.tower.healthInfo.Tau)))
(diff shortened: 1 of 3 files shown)


mdubied:
I am not too sure why the FoodRequired tests do not pass, as I did not implement them. Are they based on other health parameters?

jzzheng22:
> I am not too sure why the FoodRequired tests do not pass, as I did not implement them. Are they based on other health parameters?

They're failing because the numbers in the tests assume you don't round. Can you fix the `want` value where they are now wrong please?

mdubied:
So the want value is just the value that was manually computed?

jzzheng22:
> So the want value is just the value that was manually computed?

Yes

mdubied:
and which parameters are used? The ones in the config.json file produced by the frontend, or the ones in the config.json file that can be directly found in the first level of the repo?

jzzheng22:
> and which parameters are used? The ones in the config.json file produced by the frontend, or the ones in the config.json file that can be directly found in the first level of the repo?

I used the ones that are in `health_test.go`.

jzzheng22 (approved):
Thanks Mathieu

PR #249, opened 2022-01-06
Team5 satisfaction

UMust99:
# Summary


pkg/agents/team5/agent.go
```diff
@@ -60,17 +69,31 @@ func (a *CustomAgent5) updateSelfishness() {
...
-	if PercentageHP(a) >= 100 {
-		a.satisfaction = 3
+	tmp := int(a.CurrPlatFood())
+	if a.lastFood*120.0/100.0 <= tmp {
+		a.satisfaction += 2
+	} else if a.lastFood*110.0/100.0 <= tmp {
+		a.satisfaction += 1
+	} else if a.lastFood*70.0/100.0 >= tmp {
+		a.satisfaction -= 2
+	} else if a.lastFood*8.0/100.0 >= tmp {
```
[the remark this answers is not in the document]
UMust99: yeah you're right, it should be 80

pkg/agents/team5/agent.go
```diff
@@ -132,12 +161,14 @@ func (a *CustomAgent5) Run() {
...
 		}
+
 		a.lastMeal = lastMeal
 		if a.lastMeal > 0 {
 			a.daysSinceLastMeal = 0
 		}
-		a.updateSatisfaction()
 		a.hpAfterEating = a.HP()
-		//a.messagingCounter = 0
+		a.updateSatisfaction()
```
Rhys-J, line 170: Does this function call need to be removed?

UMust99: Yeah, satisfaction should be updated only once at the beginning of the block where the agent tries to eat so this one should be removed

pkg/agents/team5/agent.go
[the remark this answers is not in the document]
UMust99: but the order of checks is ok, if it is less than 70% of the food from previous day(line 77) than we decrease satisfaction by 2 and don't need to check if it's less than 80%

pkg/agents/team5/agent.go
Rhys-J: Thank you for removing

PR #251, opened 2022-01-06
[TEAM6] food range

gxr17:
# Summary
A function that checks for an incoming Treaty's validity. Checks to see if the Request type and Request value is valid. 

It returns min, max, bool values. Min and Max can then be used to determine how much food an agent can take. 
NOTE: The one exception is if there is no maximum value! i.e if there are only greater than request operators. In which case Max will print as 0. In this case, the agent is free to take as much as they want in theory.
True if valid. 

## Test Plan
Tested in Run function in Config.go with different treaties added.


.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
...
+    "configurations": [
+        {
+            "name": "Launch Package",
+            "type": "go",
+            "request": "launch",
+            "mode": "auto",
+            "program": "${fileDirname}"
+        }
+    ]
+}
```
MattScottEEE, line 15: Not sure what this is - shouldn't be in final repo

pkg/agents/team6/config.go
```diff
@@ -123,13 +124,44 @@ func (a *CustomAgent6) Run() {
...
+	fmt.Println(d)
+	fmt.Println(e)
 
 	// fmt.Println(a.ActiveTreaties())
+	// fmt.Println(len(a.ActiveTreaties()))
 
+	b := 0
+	for b < 100 {
+
+	}
```
MattScottEEE, line 166: Not sure what this does

pkg/agents/team6/config.go
```diff
@@ -123,13 +124,44 @@ func (a *CustomAgent6) Run() {
...
+	// treaty4.SignTreaty()
+	// a.AddTreaty(*treaty4)
+
+	//treaty5 := messages.NewTreaty(1, 1, 1, 4, 1, 4, 5, a.ID())
+	treaty5 := messages.NewTreaty(1, 1, 1, 101, 1, 1, 5, a.ID())
+
+	c, d, e := a.foodRange(*treaty5)
+	fmt.Println(c)
+	fmt.Println(d)
+	fmt.Println(e)
```
MattScottEEE, line 158: Debug code shouldn't be in final commit

pkg/agents/team6/consumption.go
```diff
@@ -77,17 +78,84 @@ func FoodRequired(currentHP int, goalHP int, healthInfo *health.HealthInfo) food
...
+func (a *CustomAgent6) foodRange(trty messages.Treaty) (int, int, bool) {
+	chkTrtyVal := trty.RequestValue()
+	mini := 0
+	maxi := 0
+	out := false
+	eqFound := 0
+	eqVal := 0
+	noMaxi := 0
+	listActTrty := a.ActiveTreaties()
+	for _, value := range listActTrty {
```
MattScottEEE, line 90:
```suggestion
	for _, value := range a.ActiveTreaties() {
```

pkg/agents/team6/consumption.go
```diff
@@ -77,17 +78,84 @@ func FoodRequired(currentHP int, goalHP int, healthInfo *health.HealthInfo) food
...
 	//if a.tower
 	//foodOnPlatform := foodOnPlatform
 	// else
 	//foodOnPlatform = lastFoodOnPlatform
 
-
 	intendedFoodIntake := a.desiredFoodIntake()
 	if a.reqLeaveFoodAmount != -1 {
-		intendedFoodIntake = min(a.tower.foodOnPlatform - a.reqLeaveFoodAmount, intendedFoodIntake)
+		//intendedFoodIntake = min(a.tower.foodOnPlatform - a.reqLeaveFoodAmount, intendedFoodIntake)
```
MattScottEEE, line 158: you'll have to resolve this merge conflict - it was fixed in Aru's/my code

MattScottEEE (changes requested):
Overall happy bar minor changes:

Not 100% sure about how `foodRange()` works, but if you've tested, then I'm good to go

pkg/agents/team6/config.go
gxr17: Just for testing. Apparently it was looping the run so printing many active treaty maps

MattScottEEE: It always loops run - this is how the agents act every tick (infra side)

pkg/agents/team6/config.go
MattScottEEE: (no fmt.Println...)

pkg/agents/team6/config.go
MattScottEEE, line 166:
```suggestion
```

pkg/agents/team6/config.go
MattScottEEE, line 158:
```suggestion
	c, d, e := a.foodRange(*treaty5)
```

MattScottEEE (changes requested):
Just stuff in run

pkg/agents/team6/consumption.go
```diff
@@ -77,17 +78,83 @@ func FoodRequired(currentHP int, goalHP int, healthInfo *health.HealthInfo) food
 	return food.FoodType(healthInfo.Tau * math.Log(healthInfo.Width/denom))
 }
 
+func (a *CustomAgent6) incomingTreatyValidityAndFoodRange(trty messages.Treaty) (int, int, bool) {
```
MattScottEEE, line 81: Can you split this into 2 functions? Sorry to be a pain - they just do quite different things (1 iterates through all treaties, the other sees if only 1 is valid). Surely foodRange doesn't need any arguments?

MattScottEEE (changes requested):
If I want to check if a treaty is valid, I don't want to have to iterate through all my existing treaties. Plz split into 2 functions

pkg/agents/team6/consumption.go
gxr17: foodRange will need the list of active treaties

MattScottEEE: yeah exactly. This is innate to the agent (so doesn't require arguments).

MattScottEEE: true, but it doesn't need the incoming treaty

MattScottEEE: just call food range inside of the valid treaty function. If I want the food range, I shouldn't have to always pass in a new treaty

MattScottEEE (approved):

PR #252, opened 2022-01-06
Add empty array checks to min and max

daryllimyt:
Self explanatory, but additionally moved back to utils because it made more sense to me to be with Average etc. Fixes -inf and inf coming up on the dashboard


priyachh (approved):
LGTM :)

PR #255, opened 2022-01-06
[APP] Added total messages sent , messages per agent graph

priyachh:
# Summary
Added a report card for total messages sent, and a graph for messages sent per agent. 

## Test Plan

Images of the additions tested below in the simulation. 

image

image

### DIFF
--- a/frontend/src/Helpers/API.ts
+++ b/frontend/src/Helpers/API.ts
@@ -6,7 +6,7 @@ import { Result } from "./Result";
 import { GetSimConfig, SimConfig } from "./SimConfig";
 
 function endpoint(req: string) {
-  return (process.env.REACT_APP_DEV ? "http://localhost:9000/" : window.location) + req;
+  return ("http://localhost:9000/") + req;
 }
 
 function parseResponse(res: any, key: string) {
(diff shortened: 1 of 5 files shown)


frontend/src/Helpers/Logging/StoryLog.ts
```diff
@@ -43,3 +43,34 @@ export function GetStoryLogs(filename: string): Promise<StoryLog[]> {
...
+              day: e["day"],
+            };
+            return m;
+          })
+        )
+      )
+      .catch((err) => reject(err));
+  });
+}
+
```
JaafarRammal, line 76: We already pull the story logs using `GetFile(filename, "story")` prior. I think the function should take an array of story logs,  check the type, and if the log type is `message`, parse it `as StoryMessageLog` and add it to the array

frontend/src/Helpers/API.ts
```diff
@@ -1,12 +1,12 @@
...
 import { DeathLog, GetDeathLogs } from "./Logging/Death";
 import { FoodLog, GetFoodLogs } from "./Logging/Food";
-import { GetStoryLogs, StoryLog } from "./Logging/StoryLog";
+import { GetStoryLogs, StoryLog, GetStoryMessageLogs, StoryMessageLog} from "./Logging/StoryLog";
 import { Result } from "./Result";
 import { GetSimConfig, SimConfig } from "./SimConfig";
 
 function endpoint(req: string) {
-  return (process.env.REACT_APP_DEV ? "http://localhost:9000/" : window.location) + req;
+  return ("http://localhost:9000/") + req;
```
JaafarRammal, line 9: Better not push dev modes :)
We now use env files, you can add a `.env` to `/frontend` and put `REACT_APP_DEV=true`

JaafarRammal (changes requested):

Issue #257, opened 2022-01-06
[APP] Add agent utility to the StatsViewer


daryllimyt:
I'm moving the main discussion points from slack here, so we don't do any accidental duplicate work

## Motivation
Pitt said in the meeting that he'd like us to be able to show each agent's utility - the issue with that is we don't have a standardised way of measuring this. 

## Approach
Utility calculations should involve the allocation of resources 
Standard utility would be a measure of the food received (each day) out of the total available food.
We have to note that utility is (in essence) similar to a transaction in that we "earn" utility (HP gain due to food) and "spend" utility (HP drain due to messaging / cost of living etc)

This gives 2 baseline approaches to modeling utility:
1. Use total food intake as a fraction of the total food available to it
2. Use "net profit" of utility = `hp gain due to food - hp loss due to CofL/comms/other`

It would make sense to then combine either of these with a mapping (utility function) such as `U(foodIntake) = log(foodIntake)` (at least an increasing, concave function). Ideally we'd have some piecewise/bounded function involving a log or a concave quadratic to account for sensitivity of obtaining further utility at various utility levels. i.e. at lower levels of utility, consuming a unit of food will grant higher utility than if this same unit of food were to be consumed at a higher level.

We would also want to normalize these by the number of deaths, to give an average utility per lifespan metric. 

## Consensus
Ultimately whichever approach we choose doesn't matter, just needs to be standardized over all agents. 
Alternative was to use HP, however it doesn't really consider resource allocation.

## Additional
This doesn't seem to have a use for agents' decision making so it's going to be just for visualization/stats purposes

JaafarRammal:
One way is:
- Implement the utility computation model in the backend
- In base agent, modify the story state to include the utility (you will also have to modify the data struct in the logger)
- In the story state logging function, use the utility variable and add it to the story log for the agent state

The story state doesn't log the agent state every tick, only when a message / death / food change happens, so if you don't want that you can add a logger in the state loggers for the utility and report the utility every tick or every day

PR #261, opened 2022-01-07
Add utility upon death and average utility per agent type

daryllimyt:
https://github.com/SOMAS2021/SOMAS2021/issues/257

4 new metrics for base agent:
- totalFoodTaken food.FoodType
- totalFoodSeen  food.FoodType
- totalHPGained  int
- totalHPLost    int

Currently, base utility is the product of these two: (please critique and add comments, this probably isn't final)
- HPGainFactor - calculated using either 
    - Ratio: gain/loss and 1 + gain/maxhp (when loss == 0)
    - Net HP: 1 + (gain - loss) / maxhp, limited to values >= 0 only
- foodIntakeFactor
    - Ratio: 0 when total seen == 0, intake / total seen otherwise

This base utility value is then passed into another function to make utility levels context sensitive i.e. higher levels less marginal utility gain etc. (but I'm not really sure if I'm doing this right, or if the previous steps make sense. Kind of just swinging blindly here). Currently I just have a quadratic function as the final utility function:
image

Frontend looks something like this:
image


jzzheng22 (commented):
LGTM. Not looking at the logging code - I didn't write that so don't feel comfortable commenting on it.

The utility code is fine.

Haven't looked at frontend either (code or UI).

Agree the utility function should probably be improved, but fine for now. (CC @MattScottEEE who will have thoughts on this)

pkg/infra/baseAgent.go
```diff
@@ -210,10 +233,12 @@ func (a *Base) TakeFood(amountOfFood food.FoodType) (food.FoodType, error) {
 	if amountOfFood < 0 {
 		return 0, &NegFoodError{}
 	}
+	a.totalFoodSeen += a.tower.currPlatFood
```
JaafarRammal, line 236: This is called:
- Once per tick
- Only if agent is calling `TakeFood`

Which means that if the agent never tried to take food we are not logging that they are seeing this food

pkg/infra/tower.go
```diff
@@ -100,6 +100,7 @@ func (t *Tower) endOfDay() {
 	t.dayInfo.CurrDay += 1
 	for _, agent := range t.Agents {
 		agent := agent.BaseAgent()
+		t.stateLog.LogUtility(t.dayInfo, agent.agentType, agent.Utility(), agent.IsAlive())
```
JaafarRammal, line 103: I think `IsAlive` is updated after hp decay I called on the line after

pkg/infra/baseAgent.go
```diff
@@ -226,6 +251,77 @@ func (a *Base) TakeFood(amountOfFood food.FoodType) (food.FoodType, error) {
 	return foodTaken, nil
 }
 
+func (a *Base) Utility() float64 {
```
JaafarRammal, line 254: I'd prefer lower case `utility` since this is only called in the same package by tower

JaafarRammal (commented):
Frontend code LGTM
Logging code LGTM
I think on the backend agent death check is being called too early?

pkg/infra/tower.go
daryllimyt: I see I see, will change

pkg/infra/baseAgent.go
daryllimyt: sure np

pkg/infra/baseAgent.go
daryllimyt: I've moved this into a function that gets called on each tick and occurs before each agent takes action

JaafarRammal (dismissed):

JaafarRammal (approved):

PR #262, opened 2022-01-07
Add a parameter MaxFoodIntake

mdubied:
# Summary

This parameter is contained in the HealthInfo struct and limits the amount of food that an agent can physically eat. Choosing MaxFoodIntake large enough (4*tau=60 for example), we let the possibility for the agent to eat way more than needed, but still insure that they cannot eat the total amount of food on the platform (if curPlatFood>MaxFoodIntake)


mdubied:
It seems that there is a problem on the frontend side. @udai-arneja can you have a closer look?

udai-arneja:
I wrote some of the code so won't give a review, will wait for Jason

hussain2603 (dismissed):

hussain2603 (commented):
someone else should probs review frontend since I don't know that bit

jzzheng22 (commented):
Code looks fine.

Shouldn't we make `MaxFoodIntake` dependent on `tau` though?

mdubied:
We can, currently we can freely choose it on frontend. We chose the default paramters as 4*tau, where tau takes the best tuning we have so far

jzzheng22:
> We can, currently we can freely choose it on frontend. We chose the default paramters as 4*tau, where tau takes the best tuning we have so far

I feel like it's better to just set MaxFoodIntake to always be 4*tau is all - what happens if someone changes tau but then doesn't change the maxFoodIntake?

mdubied:
Yes, I am also happy with this solution

udai-arneja:
We cant change tau from the frontend, I dont think we want to change it at all.

frontend/src/Components/NewRunConfig/Settings/Settings.tsx
```diff
@@ -21,6 +21,7 @@ export default function Settings(props: SettingsInterface) {
     var key: Key = keyString; // converting keyString to type Key
     config[key] = value;
     setConfig(config);
+    console.log(config);
```
JaafarRammal, line 24: I think a print that should be removed!

pkg/infra/baseAgent.go
```diff
@@ -210,7 +210,7 @@ func (a *Base) TakeFood(amountOfFood food.FoodType) (food.FoodType, error) {
 	if amountOfFood < 0 {
 		return 0, &NegFoodError{}
 	}
-	foodTaken := food.FoodType(math.Min(float64(a.tower.currPlatFood), float64(amountOfFood)))
+	foodTaken := food.FoodType(utilFunctions.MinInt(int(a.tower.currPlatFood), int(amountOfFood), int(a.tower.healthInfo.MaxFoodIntake)))
```
JaafarRammal, line 213: It means that agents are not allowed to destroy the food anymore (which is what selfish and random agents were doing).

JaafarRammal (dismissed):
Code LGTM, left 2 minor comments

jzzheng22:
> We cant change tau from the frontend, I dont think we want to change it at all. So you can change maxFoodIntake, but not tau. @jzzheng22

If the max food intake is based on tau, and you can't change tau, then surely max food intake shouldn't be changable?

udai-arneja:
I think, while maxFood intake should typically be 4*tau, it would be interesting to see behaviour with the possibility of 'destroying food' vs not being able to destroy it... open to opinions @jzzheng22 @mdubied @JaafarRammal

pkg/infra/baseAgent.go
udai-arneja: Yep, unless we let them destroy food by setting the food intake to 100 or whatever the max is.... think this is a good point for experimentation... Open to opinions tho

frontend/src/Components/NewRunConfig/Settings/Settings.tsx
udai-arneja: Indeed my bad haahah

mdubied:
I think we can leave it as the PR proposes, i.e. as a parameter on the frontend. We will probably run most of our simulations with the default value 4 tau, but as Udai mentioned, it can also be interesting to run simulations where maxFoodIntake="maxFoodOnPlatform"

jzzheng22:
Happy with that then, fair enough.

jzzheng22 (approved):

frontend/src/Components/NewRunConfig/Settings/Settings.tsx
mdubied: print removed

JaafarRammal (approved):

PR #264, opened 2022-01-08
Custom Logging for Team 2 CSV

JaafarRammal:
# Summary

Team 2 needs to export some CSV files at the end to evaluate their agents
I added a `CustomLogs` function to base agent that gets called at the end of the simulation only if you set flag `--log` to your agent type.

E.g.: `go run main.go --log=Team1Agent1`

I also fixed the config file because it seems like we allowed an old config file version to go through in the commits with the wrong key names


jzzheng22 (approved):
LGTM but haven't tested

JaafarRammal:
image
I hope this counts as testing :o

PR #266, opened 2022-01-08
[Team 5] Update favour calculation

tchiko7:
# Summary

added favour calculation for neighbouring agents

### DIFF
--- a/pkg/agents/team5/agent.go
+++ b/pkg/agents/team5/agent.go
@@ -35,7 +35,7 @@ type CustomAgent5 struct {
 	attemptToEat      bool
 	leadership        int
 	lastSeenFood      food.FoodType
-	sensitivity       int
+	sensitivity       float64
 	avgPlatFood       food.FoodType
 	daysSinceShuffle  int
 	// Social network of other agents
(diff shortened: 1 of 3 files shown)


Rhys-J:
NOTES: 
-Check what change is bounded by, possibly implement restrictions of how much favour can change by
-Why only judge the agent directly above and below when we gather information via messaging from those on +-2 floors.
-Remove old Update Favour and the calls to that function
-Clarification of FoodExp, does is mean different things in each function, if so, change name as it is confusing
-Are there any ways to use satisfaction in these calculations.

pkg/agents/team5/social.go
```diff
@@ -72,6 +72,78 @@ func (a *CustomAgent5) updateFavour() {
...
+		foodTotal += a.socialMemory[uuid].foodTaken
+	}
+	foodExp := float64(foodTotal) / float64(len(a.surroundingAgents)+1)
+	if foodExp == 0 {
+		foodExp = float64(a.attemptFood) // to avoid favour shooting extremely high
+	}
+	return foodExp
+}
+
+func (a *CustomAgent5) calculateFavourBelow(belowID uuid.UUID) (int, food.FoodType) {
```
daryllimyt, line 109: Is there a way to combine the favor calculation to use 1 function? Seems like there's some redundancy

pkg/agents/team5/agent.go
```diff
@@ -151,6 +164,8 @@ func (a *CustomAgent5) Run() {
 		a.treatyOverride()
 		if a.Floor() == a.rememberFloor { // if the agent is still on the same floor it can update its satisfaction
```
Rhys-J, line 165: remeberFloor is updated in dayPassed, so these values should always be equal at this point in the run() code

pkg/agents/team5/social.go
```diff
@@ -72,6 +72,78 @@ func (a *CustomAgent5) updateFavour() {
...
+	sens := a.sensitivity
+	foodIntent := float64(a.attemptFood)
+	hpSelf := float64(a.HP())
+
+	maxHP := float64(a.HealthInfo().MaxHP)
+
+	hpBelow := float64(a.socialMemory[belowID].agentHP)
+	foodBelow := float64(a.socialMemory[belowID].foodTaken)
```
Rhys-J, line 117: We call these functions when the platform is on our floor, so the agent below would not have eaten that day. Is this acceptable that this will use the food they ate in the previous day?

tchiko7: its acceptable cause the favour change due to their hp and foodtaken is small and will only lag behind by 1 day

tchiko7: but if its possible to do like you say then its probably better

jzzheng22:
Superceded by #275

PR #267, opened 2022-01-08
[TEAM6] Trust

MattScottEEE:
# Summary

Added variables for trust and means of keeping track of neighbours. Will continue with implementing trust in parallel with messages in future

## Additional Information

Changed initialisation of array in config:

`memory{}` looks like an empty struct of type memory. `make(memory, 0)` clarifies that it's a slice (and calls `runtime.makeslice`)

neighbours struct stores the conceived IDs of the above and below neighbours for use in the trust map


mdubied:
Good To Go

mdubied (approved):
LGTM

PR #270, opened 2022-01-09
Improve utility calculation

daryllimyt:
## Motivation
Current state of utility is a bit weird - on testing random agent reaches the most amount of utility, probably due to weighting of food intake variable, where random agent on average (probably?) takes the most food of all agents. 

That said, the calculation does need some sort of revamp but not sure how exactly.

## Changes
Update utility to just use a weighted sum of various factors (netHPGainFactor, foodIntakeFactor) with sum of weights = 1
0.75*foodIntakeFactor + 0.25*netHPGainFactor

Overall the utility seems more balanced all teams' agents
image

### DIFF
--- a/pkg/infra/baseAgent.go
+++ b/pkg/infra/baseAgent.go
@@ -271,7 +271,8 @@ func (a *Base) utility() float64 {
 	hpGainFactor := netHPGainFactor(a.totalHPGained, a.totalHPLost, a.tower.healthInfo.MaxHP)
 	foodIntakeFactor := foodIntakeFactor(a.totalFoodTaken, a.totalFoodSeen)
 	// this means the multiplcation of these values can be larger than 1
-	return simpleQuadraticUtility(hpGainFactor * foodIntakeFactor)
+	// return simpleQuadraticUtility(hpGainFactor * foodIntakeFactor)
+	return 0.75*foodIntakeFactor + 0.25*hpGainFactor
 }
 
 /*
@@ -289,12 +290,6 @@ key values:
 - > 1: agent gaining hp on average
 - when loss is 0, default to percentage difference between gain and loss. Acknowledge that HP gain should increase utility
 */
-// func hpGainFactor(gain, loss, maxhp int) float64 {
-// 	if loss == 0 {
-// 		return 1 + float64(gain)/float64(maxhp)
-// 	}
-// 	return float64(gain) / float64(loss)
-// }
 
 /*
 Net HP Gain Factor.
@@ -322,16 +317,16 @@ func foodIntakeFactor(intake, seen food.FoodType) float64 {
 	return float64(intake) / float64(seen)
 }
 
-// Quadratic function standard form: a(x-h)^2+k
-func quadratic(x, a, h, k float64) float64 {
-	return a*math.Pow((x-h), 2) + k
-}
+// // Quadratic function standard form: a(x-h)^2+k
+// func quadratic(x, a, h, k float64) float64 {
+// 	return a*math.Pow((x-h), 2) + k
+// }
 
-// Utility function is a strictly increasing function that starts at 0,0 ends at 1,1
-func simpleQuadraticUtility(x float64) float64 {
-	// -(x-1)^2 + 1
-	return utilFunctions.RestrictToRange(0, 1, quadratic(x, -1, 1, 1))
-}
+// // Utility function is a strictly increasing function that starts at 0,0 ends at 1,1
+// func simpleQuadraticUtility(x float64) float64 {
+// 	// -(x-1)^2 + 1
+// 	return utilFunctions.RestrictToRange(0, 1, quadratic(x, -1, 1, 1))
+// }
 
 func (a *Base) HealthInfo() *health.HealthInfo {
 	return a.tower.healthInfo


hussain2603:
I think this calculation doesn't make any sense unless we take into account number of agent type deaths/max surviving age of agent type. This is also only for individual utility, I think we should also have social utility where we try to see if agents are pushing for overall survival (not entirely sure how this could be calculated). 

What does everyone think @SOMAS2021/team-leads ?

MattScottEEE:
Think it would be nice to reiterate the point I made on Slack:

An agent who takes all the food will have a very high "utility" and small number of deaths, seeming like a good performance. The issue is how this relates to global utility, as it is necessary that this detracts from the utility of the rest of the tower and results in overall lower utility and greater deaths

We propose an example of a purely collectivistic agent, who takes the bare minimum for survival. This means that no utility is wasted, as it is purely used for self-sufficiency, with no impact on the surrounding agents. Resultantly, we offer this as a new metric:

utility = (average food requested by tower + 1)  /  (food received + 1)  (no div 0!) TODO: some scaling by global deaths 

A purely collectivistic agent will ask for 2 food every 3 turns (to not die and teeter around the limit for critical death days) and receive 2 food, compared to a global average of 2 food (assuming that we have an equally collectivistic society). This makes the system stable (check team6 trial with purely collectivist agents, we can have 100 with 0 deaths). This is true, 100% utility. Nothing is wasted and nothing is lost.

Contrast this with a purely evil agent - they will request the whole platform each time (and receive it too, at the dismay of the neighbours), resulting in a lot more deaths. The issue is, the rest of the tower are fine with just a little bit, so you're massively wasting resources relative to the other agents.


Again, this is speculative. Need other opinions. Hope this is at least interesting

mdubied:
I would like to propose a _global_ utility function, measuring the _global_ "welfare" of the group of agents:

`Utility = Sum over all agents (FoodReceived/FoodRequested) / # agents`

Every agents can contribute to a _part-utility_ of 1 (maximum value of FoodReceived/FoodRequested). As we sum over all agents, the global utility can have a maximum value of # agents, so that we divide by # agents to get Utility in [0,1]. If the FoodRequest by an agent is 0, then its part-utility is defined as 1 (special case, as we have mathematically 0/0)

_2 Examples_¨
A) The tower consists of collectivist agents, that only take food when really needed, and they need the exact amount needed to survive. According to the specs, this scenario is possible. In this case, every agent part-utility is 1, and the global utility achieves its maximum value of 1
B) We have some greedy agents mixed to the collectivist agent of case A). If a greedy agent get all the food he requested,  he reaches a part-utlity of 1. However, the other agents are unlikely to get the food they asked for, also if this amount is small (there is not enough food on the platform for everyone, as the greedy agent took a large amount of it). Their part-utility is 0 for the ones that cannot eat. The total utility is also <1.
This way, we see how the greedy agent "wastes" the common pool resource. This leads to a lower global welfare of the society he is living in.

This is just a proposition, but it seems interesting to me :)

MattScottEEE:
> I would like to propose a _global_ utility function, measuring the _global_ "welfare" of the group of agents:
> 
> `Utility = Sum over all agents (FoodReceived/FoodRequested) / # agents`
> 
> Every agents can contribute to a _part-utility_ of 1 (maximum value of FoodReceived/FoodRequested). As we sum over all agents, the global utility can have a maximum value of # agents, so that we divide by # agents to get Utility in [0,1]. If the FoodRequest by an agent is 0, then its part-utility is defined as 1 (special case, as we have mathematically 0/0)
> 
> **2 Examples** **A)** The tower consists of collectivist agents, that only take food when really needed, and they need the exact amount needed to survive. According to the specs, this scenario is possible. In this case, every agent part-utility is 1, and the global utility achieves its maximum value of 1 **B)** We have some greedy agents mixed to the collectivist agent of case A). If a greedy agent get all the food he requested, he reaches a part-utility of 1. However, the other agents are unlikely to get the food they asked for, also if this amount is small (there is not enough food on the platform for everyone, as the greedy agent took a large amount of it). Their part-utility is 0 for the ones that cannot eat. The total utility is also <1. This way, we see how the greedy agent "wastes" the common pool resource. This leads to a lower global welfare of the society he is living in.
> 
> This is just a proposition, but it seems interesting to me :)

I suppose it really isn't possible to discount the greed of a narcissist - the truth is that they will always have maximal utility, because they sabotage the group. Given this, I love this proposition - we can even take local utility of an agent to be `receive/request`

Correct me if I'm wrong, but this should also implicitly scale with deaths:

Having more deaths = less individual utility = less global utility

LGTM?

mdubied:
@MattScottEEE is `agent type` in your formula corresponding to the different groups agents? Like team6, team 1 ...?

MattScottEEE:
> @MattScottEEE is `agent type` in your formula corresponding to the different groups agents? Like team6, team 1 ...?

Yes, as then the total utility is 

Sum over teams (average utility per team * number of agents per team) / total agents

MattScottEEE:
IMPORTANT - FINAL ANSWER ON UTILITY (SUPPORTED BY PITT)

For this paper, we take N agents i ∈{1,...,N}that perform the following actions in each iterated round t ∈{1,...,∞}:
1. Determines the resources it has available, gi ∈[0,1]
2. Determines its need for resources, qi ∈[0,1]
3. Makes a provision of resources, pi ∈[0,1]
4. Makes a demand for resources, di ∈[0,1]
5. Receives an allocation of resources, ri ∈[0,1]
6. Makes an appropriation of resources, r′i ∈[0,1]
The total resources accrued at the end of a round is hence given by:
Ri = r′i + (gi −pi) (1)
Which provides the utility of an agent, ui, as:
ui =
αi(qi) + βi(Ri −qi), if Ri ≥qi
αi(Ri) −γi(qi −Ri), otherwise 

In this equation, αi, βi and γi are private coefficients measuring, respectively, the relative utilities of getting resources that are needed, getting resources that are not needed, and not getting resources that are needed.

MattScottEEE:
Superseded by #297

PR #271, opened 2022-01-09
[APP] Fix memory, make story optional

JaafarRammal:
# Summary

- Added optional tick filter at the backend /read endpoint
- Change on frontend how we read a story
- Make logging the story optional
- Reduce food log size


jzzheng22 (approved):
Looks okay from a quick skim. ideally return errors instead of -1 but fine since we are close to deadline.

Can we still display when an agent dies?

jzzheng22 (approved):

PR #273, opened 2022-01-09
[TEAM 6] Treaties

ap8718:
# Summary
Implemented treaty proposal and evaluation, and calls these functions once a day.


MattScottEEE (approved):
Fine - remember to return original JSON after merge to main

PR #285, opened 2022-01-09
[TEAM6] Final PR

MattScottEEE:
# Summary

Added treaty strategies, messaging + behavioural changes

## Test Plan

Rigorous testing in run - agents are fair and honest


config.json
```diff
@@ -1,6 +1,6 @@
 {
     "UseFoodPerAgentRatio": false,
-    "FoodOnPlatform": 100,
+    "FoodOnPlatform": 10,
```
JaafarRammal, line 3: Prefer to leave config as is!

pkg/agents/team6/behaviours.go
```diff
@@ -4,46 +4,52 @@ import (
...
+	hpScore := 1 - float64(a.HP())/float64(a.HealthInfo().MaxHP)
 	floor := a.Floor() + 1
 
+	// Update agent's guess of how many floors there are in the tower
 	if floor > a.maxFloorGuess {
 		a.maxFloorGuess = floor + 1
 	}
 
+	// Utility func for floor
 	fdNum := float64(floor) / float64(a.maxFloorGuess) * aConf.lambda
```
JaafarRammal, line 24: Just checking that `float64(a.maxFloorGuess)` is never 0?

pkg/agents/team6/treaties.go
```diff
@@ -0,0 +1,375 @@
...
+	// Calculates the function scaling parameter a
+
+	var result float64
+
+	if socialMotive == "Altruist" || socialMotive == "Nacissist" {
+		// Don't scale depending on desired food
+		result = params.g*math.Pow(x, 1/params.r) - params.c*x
+	} else {
+		a := (1 / z) * math.Pow((params.c*params.r)/params.g, params.r/(1-params.r))
+		result = params.g*math.Pow(a*x, 1/params.r) - params.c*a*x
```
JaafarRammal, line 30: I see you set `params.r` to never be 0 in the other function, but just confirming that `r` will never be set to 0 or 1 during sim?

JaafarRammal (changes requested):
Code LGTM, a few comments left.
Would be good to revert config file.
I think `func (a *CustomAgent6) considerTreaty(t *messages.Treaty) bool` in `pkg/agents/team6/treaties.go` can be visually improved as it's very nested

pkg/agents/team6/communication.go
```diff
@@ -0,0 +1,254 @@
...
+	case "Altruist":
+		reply = true
+
+	case "Collectivist":
+		if currentHP >= levels.weakLevel {
+			reply = true
+		} else {
+			reply = false
+		}
+
```
jzzheng22, line 109:
```suggestion
		reply = currentHP >= levels.weakLevel
```

pkg/agents/team6/communication.go
```diff
@@ -0,0 +1,254 @@
...
+	case "Selfish":
+		if currentHP >= levels.strongLevel {
+			reply = true
+		} else {
+			reply = false
+		}
+
```
jzzheng22, line 116:
```suggestion
		reply = currentHP >= levels.strongLevel
```

pkg/agents/team6/behaviours.go
MattScottEEE: Nice spot - we always initialise as our current floor and then update if we get a score bigger - not possible to get 0

config.json
MattScottEEE: Yes that's on me - I thought I updated from main but perhaps not!

pkg/agents/team6/communication.go
```diff
@@ -0,0 +1,254 @@
...
+	if reply {
+		a.reqLeaveFoodAmount = msg.Request()
+		a.Log("I received a requestTakeFood message and my response was true")
+	} else {
+		a.reqLeaveFoodAmount = -1
+		a.Log("I received a requestTakeFood message and my response was false")
+	}
+	// Try to identify our neighbours
+	a.identifyNeighbours(msg.SenderID(), msg.SenderFloor())
```
jzzheng22, line 154: ??

Dead code
Just pass `false` directly into `Reply()`
And you don't need the if statement, since the if block will never execute, only the else

pkg/agents/team6/treaties.go
MattScottEEE: Again, fair spot - these params are static (and we have a special case for when we have `params.r == 1`)

pkg/agents/team6/config.go
```diff
@@ -59,19 +107,131 @@ func New(baseAgent *infra.Base) (infra.Agent, error) {
...
-		maxFloorGuess: baseAgent.Floor() + 2,
-		foodTakeDay:   0,
+		currBehaviour:       initialBehaviour,
+		maxFloorGuess:       baseAgent.Floor() + 2,
+		foodTakeDay:         0,
+		reqLeaveFoodAmount:  -1,
+		lastFoodTaken:       0,
+		averageFoodIntake:   0.0,
+		longTermMemory:      make(memory, 0), //memory{},
+		shortTermMemory:     make(memory, 0), //memory{},
```
jzzheng22, line 125:
```suggestion
		longTermMemory:      make(memory, 0),
		shortTermMemory:     make(memory, 0),
```

pkg/agents/team6/behaviours.go
JaafarRammal: Cool

MattScottEEE:
> Code LGTM, a few comments left.
> Would be good to revert config file.
> I think `func (a *CustomAgent6) considerTreaty(t *messages.Treaty) bool` in `pkg/agents/team6/treaties.go` can be visually improved as it's very nested

Thanks for this - reverting the config + taking a look at this function:

overall, team think that, despite the nested-ness, it preserves readability instead of adding lots of sub-functions. Could I ask for a recommendation of what we might be able to do?

pkg/agents/team6/consumption.go
```diff
@@ -0,0 +1,123 @@
...
+func (a *CustomAgent6) intendedFoodIntake() food.FoodType {
+
+	desiredFoodIntake := a.desiredFoodIntake()
+	maxAllowedFood := a.maxAllowedFood()
+
+	if maxAllowedFood >= desiredFoodIntake {
+		return desiredFoodIntake
+	} else {
+		return maxAllowedFood
+	}
```
jzzheng22, line 116:
```suggestion
	}
	return maxAllowedFood
```

pkg/agents/team6/treaties.go
JaafarRammal: What I'm seeing is that if `!(socialMotive == "Altruist" || socialMotive == "Nacissist)"`, regardless of the value of `r`, this is reached. If `r` is set to 1 what is the behavior? (because you get `params.r/(1-params.r) = 1/0`)

In all fairness, the only time the calculation would be valid is when `(params.c*params.r)/params.g` is 1, you then get `x^(x/(1-x))`

If `r` never reaches an illegal value, feel free to resolve the convo!

pkg/agents/team6/memory.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+}
+
+// Returns whether agent finds themselves reassigned
+// Could be the case that reassignment has occurred, but agent remains on the same floor, so in their POV they have not been reassigned
+func (a *CustomAgent6) isReassigned() bool {
+	if a.prevFloor != a.Floor() && a.prevFloor != -1 {
+		return true
+	}
+	return false
```
jzzheng22, line 29:
```suggestion
	return a.prevFloor != a.Floor() && a.prevFloor != -1
```

pkg/agents/team6/treaties.go
```diff
@@ -0,0 +1,375 @@
...
+		strongLevel:  a.HealthInfo().MaxHP * 3 / 5,
+		healthyLevel: a.HealthInfo().MaxHP * 3 / 10,
+		weakLevel:    a.HealthInfo().WeakLevel,
+		critLevel:    0,
+	}
+	// We only consider meaningful LeaveAmountFood and LeavePercentFood treaties, i.e., the treaties with RequestOp GE or GT
+	if t.RequestOp() == messages.GE || t.RequestOp() == messages.GT {
+		switch t.Condition() {
+		// HP
+		case 1:
```
jzzheng22, line 94: Don't compare the numbers - use the enum here instead.

pkg/agents/team6/treaties.go
```diff
@@ -0,0 +1,375 @@
...
+					return a.considerTreatyUsingUtility(t)
+				} else {
+					if t.ConditionValue() >= a.HealthInfo().WeakLevel {
+						return true
+					} else {
+						a.Log("I used considerTreatyUsingUtility", infra.Fields{"my floor:": a.Floor(), "my social motive": a.currBehaviour.string()})
+						return a.considerTreatyUsingUtility(t)
+					}
+				}
+
```
jzzheng22, line 110: I suggest putting this into a function

pkg/agents/team6/treaties.go
```diff
@@ -0,0 +1,375 @@
...
+					return a.considerTreatyUsingUtility(t)
+				} else {
+					if t.ConditionValue() >= levels.strongLevel {
+						return true
+					} else {
+						a.Log("I used considerTreatyUsingUtility", infra.Fields{"my floor:": a.Floor(), "my social motive": a.currBehaviour.string()})
+						return a.considerTreatyUsingUtility(t)
+
+					}
+				}
```
jzzheng22, line 123: I suggest putting this into a function

pkg/agents/team6/treaties.go
```diff
@@ -0,0 +1,375 @@
...
+					a.Log("I used considerTreatyUsingUtility", infra.Fields{"my floor:": a.Floor(), "my social motive": a.currBehaviour.string()})
+					return a.considerTreatyUsingUtility(t)
+				} else {
+					if a.convertToTakeFoodAmount(float64(t.ConditionValue()), t.Request(), t.RequestValue()) <= 2 { // 2 is the amount needed to y sufficient to go from critical to WeakLevel
+						a.Log("I used considerTreatyUsingUtility", infra.Fields{"my floor:": a.Floor(), "my social motive": a.currBehaviour.string()})
+						return a.considerTreatyUsingUtility(t)
+					} else {
+						return true
+					}
+				}
```
jzzheng22, line 150: I suggest putting this into a function

pkg/agents/team6/treaties.go
```diff
@@ -0,0 +1,375 @@
...
+					a.Log("I used considerTreatyUsingUtility", infra.Fields{"my floor:": a.Floor(), "my social motive": a.currBehaviour.string()})
+					return a.considerTreatyUsingUtility(t)
+				} else {
+					if a.convertToTakeFoodAmount(float64(t.ConditionValue()), t.Request(), t.RequestValue()) <= 60 { // change to a.HealthInfo().MaxFoodIntake
+						a.Log("I used considerTreatyUsingUtility", infra.Fields{"my floor:": a.Floor(), "my social motive": a.currBehaviour.string()})
+						return a.considerTreatyUsingUtility(t)
+					} else {
+						return true
+					}
+				}
```
jzzheng22, line 162: I suggest putting this into a function

pkg/agents/team6/treaties.go
```diff
@@ -0,0 +1,375 @@
...
+			}
+
+		default:
+			a.Log("I used considerTreatyUsingUtility", infra.Fields{"my floor:": a.Floor(), "my social motive": a.currBehaviour.string()})
+			return a.considerTreatyUsingUtility(t)
+		}
+	} else {
+		return false
+	}
+}
```
jzzheng22, line 179: Function is too long - use functions to shorten it

config.json
```diff
@@ -1,32 +1,32 @@
...
-    "hpCritical": 3,
-    "maxDayCritical": 3,
-    "HPLossBase": 5,
-    "HPLossSlope": 0.2,
-    "LogMain": true,
-    "LogStory": true,
-    "SimTimeoutSeconds": 60
+  "UseFoodPerAgentRatio": false,
+  "FoodOnPlatform": 100,
+  "FoodPerAgentRatio": 0,
```
jzzheng22: Did yuo change this file...?

pkg/agents/team6/treaties.go
```diff
@@ -0,0 +1,375 @@
...
+		}
+
+		// 3. If we benefit from the treaty, accept it
+		return benefit > 0.0
+
+	} else {
+		// We do not consider other type of treaties than LeaveAmountFood and LeavePercentFood
+		return false
+	}
+}
```
jzzheng22, line 252: Don't need else, you can just return false

pkg/agents/team6/treaties.go
```diff
@@ -0,0 +1,375 @@
...
+			return a.HP() == t.ConditionValue()
+		// LE
+		case 4:
+			return a.HP() <= t.ConditionValue()
+		// LT
+		case 5:
+			return a.HP() < t.ConditionValue()
+		default:
+			return true
+		}
```
jzzheng22, line 278: Unsafe to use numbers, use the enum

jzzheng22: I'd also put this into a function

jzzheng22: Both of the above comments apply to the other case statements too

jzzheng22 (changes requested):
Some of the code is unsafe - needs to be changed.

pkg/agents/team6/treaties.go
MattScottEEE: this is because the value of r is specific to the behaviour (initialised in config!) - testing for behaviour also tests for r :)

config.json
JaafarRammal: Latest file on main is:

```json
{
    "UseFoodPerAgentRatio": false,
    "FoodOnPlatform": 100,
    "FoodPerAgentRatio": 0,
    "Team1Agents": 2,
    "Team1Agents2": 2,
    "Team2Agents": 2,
    "Team3Agents": 2,
    "Team4Agents": 2,
    "Team5Agents": 2,
    "Team6Agents": 2,
    "Team7Agents": 2,
    "RandomAgents": 2,
    "AgentHP": 100,
    "AgentsPerFloor": 1,
    "TicksPerFloor": 10,
    "SimDays": 8,
    "ReshuffleDays": 1,
    "maxHP": 100,
    "weakLevel": 10,
    "width": 48,
    "tau": 15,
    "MaxFoodIntake": 60,
    "hpReqCToW": 5,
    "hpCritical": 3,
    "maxDayCritical": 3,
    "HPLossBase": 5,
    "HPLossSlope": 0.2,
    "LogMain": true,
    "LogStory": true,
    "SimTimeoutSeconds": 60
}
```

pkg/agents/team6/treaties.go
MattScottEEE: (reopening so you can see response)

mdubied: `r=1` is only the case for `socialMotive="Altruist"`. In this case, we go in the first `if` statement where we do not divide by `(1-r)`

pkg/agents/team6/communication.go
```diff
@@ -0,0 +1,254 @@
...
+}
+
+// Handles the responses from other agents to our treaty proposals
+func (a *CustomAgent6) HandleTreatyResponse(msg messages.TreatyResponseMessage) {
+	if msg.Response() {
+		// Adds accepted treaty in the active treaties of the agent
+		treaty := a.proposedTreaties[msg.TreatyID()]
+		a.AddTreaty(treaty)
```
SaraFFdez, line 215: Should you sign the treaty here too to increase the treaty count?

JaafarRammal:
> overall, team think that, despite the nested-ness, it preserves readability instead of adding lots of sub-functions. Could I ask for a recommendation of what we might be able to do?

@MattScottEEE That's correct, it's actually readable, and I think it's the logs that got in the way. This was written 13 times (exactly same line), could a call wrapped to make it cleaner actually: 

`a.Log("I used considerTreatyUsingUtility", infra.Fields{"my floor:": a.Floor(), "my social motive": a.currBehaviour.string()})`

pkg/agents/team6/treaties.go
AdrianKoch3010: It is like a lookup table, putting the individual switch cases into the function is not going to improve readability

pkg/agents/team6/communication.go
MattScottEEE: Let me check this one - cheers for spotting <3

pkg/agents/team6/communication.go
MattScottEEE: False was a placeholder - updating now (`sorry for this :((( `)

pkg/agents/team6/treaties.go
jzzheng22: No, this is extremely difficult to read. 
The way you handle `Collectivist`, and `Selfish` should be in functions.
The way you handle `HP` and `AvailableFood` should be in functions.

pkg/agents/team6/communication.go
MattScottEEE: We interpreted it two ways:

1.) either we sign a treaty then sign it off
2) or we receive a treaty and sign on their behalf

We assume that the former is more logical, assuming that the treaty response we receive is already signed

MattScottEEE: (in which case, we shouldn't have to sign)

MattScottEEE:
> > overall, team think that, despite the nested-ness, it preserves readability instead of adding lots of sub-functions. Could I ask for a recommendation of what we might be able to do?
> 
> @MattScottEEE That's correct, it's actually readable, and I think it's the logs that got in the way. This was written 13 times (exactly same line), could a call wrapped to make it cleaner actually:
> 
> ```go
> a.Log("I used considerTreatyUsingUtility", infra.Fields{"my floor:": a.Floor(), "my social motive": a.currBehaviour.string()})
> ```
> 
> Maybe it should go inside the `considerTreatyUsingUtility` function since this is when you are calling this log

Great suggestion - thanks for this

pkg/agents/team6/communication.go
SaraFFdez: Fair

pkg/agents/team6/treaties.go
jzzheng22: This is largely resolved by #289

pkg/agents/team6/treaties.go
jzzheng22: This is largely resolved by #289

pkg/agents/team6/communication.go
mdubied: We updated it. It does not always return false now

pkg/agents/team6/treaties.go
```diff
@@ -0,0 +1,304 @@
...
+	default:
+		return consultUtility
+	}
+}
+
+// Decides if to accept or reject a treaty using utility. Used in "considerUtility" function above
+func (a *CustomAgent6) considerTreatyUsingUtility(t *messages.Treaty) bool {
+
+	// Log
+
```
jzzheng22, line 154:
```suggestion
```

pkg/agents/team6/treaties.go
```diff
@@ -0,0 +1,304 @@
...
+		currentShortTermUtility := a.evaluateUtility(a.shortTermMemory) // estimated utility on the current floor
+		currentLongTermUtility := a.evaluateUtility(a.longTermMemory)   // estimated utility over the entire time in the tower
+
+		// benefit of signing the treaty (> 0 == beneficial)
+		shortTermBenefit := treatyUtility - currentShortTermUtility
+		longTermBenefit := treatyUtility - currentLongTermUtility
+
+		estimatedPeriod := int(a.reassignPeriodGuess)
+		estimatedTimeLeft := estimatedPeriod - len(a.shortTermMemory)
+		benefit := 0.0
```
jzzheng22, line 190: Would personally move the long term benefit stuff into the else statement since you don't need it unless you enter the else.

pkg/agents/team6/treaties.go
```diff
@@ -0,0 +1,304 @@
...
+
+		} else {
+			// → The longer the duration of a treaty, the more important is it's long term benefit
+			// E.g.
+			// (1) time left on this level = 4, duration = 5 --> shortTermFocus 80%
+			// 		--> The treaty mostly matters in the short term
+			// (2) time left on this level = 4, duration = 8 --> shortTermFocus 50%
+			// 		--> The treaty counts as much on this level as for the time after
+			// (3) time left on this level = 1, duration = 100 --> shortTermFocus 1%
+			// 		--> The treaty mostly matters in the long term
```
jzzheng22, line 210: optional, but maybe consider whether these comments are necessary here

pkg/agents/team6/treaties.go
```diff
@@ -0,0 +1,304 @@
...
+		critLevel:    0,
+	}
+	proposedTreaty := messages.NewTreaty(1, a.HealthInfo().MaxHP, 2, 1, 1, 1, int(2*a.reassignPeriodGuess), a.ID())
+	// Altruist and Narcissist do not propose treaties
+	switch a.currBehaviour.string() {
+	case "Collectivist":
+		switch rand.Intn(2) {
+		case 0:
+			// ConditionType, conditionValue, RequestType, requestValue, cop, rop, duration, proposerID
+			proposedTreaty = messages.NewTreaty(1, a.HealthInfo().WeakLevel, 2, 1, 1, 1, int(2*a.reassignPeriodGuess), a.ID())
```
jzzheng22, line 271: Up to you here, but I would use the ConditionType and RequestType, not the numbers

pkg/agents/team6/treaties.go
```diff
@@ -0,0 +1,304 @@
...
+		case 1:
+			// Different treaty
+			proposedTreaty = messages.NewTreaty(1, a.HealthInfo().WeakLevel, 2, 1, 1, 1, int(4*a.reassignPeriodGuess), a.ID())
```
jzzheng22, line 274: Up to you here, but I would use the ConditionType and RequestType, not the numbers

pkg/agents/team6/treaties.go
```diff
@@ -0,0 +1,304 @@
...
+		}
+
+	case "Selfish":
+		switch rand.Intn(2) {
+		case 0:
+			// ConditionType, conditionValue, RequestType, requestValue, cop, rop, duration, proposerID
+			proposedTreaty = messages.NewTreaty(1, levels.strongLevel, 2, 1, 1, 1, int(2*a.reassignPeriodGuess), a.ID())
```
jzzheng22, line 281: Up to you here, but I would use the ConditionType and RequestType, not the numbers

pkg/agents/team6/treaties.go
```diff
@@ -0,0 +1,304 @@
...
+		case 1:
+			// Different treaty
+			proposedTreaty = messages.NewTreaty(1, levels.strongLevel, 2, 1, 1, 1, int(4*a.reassignPeriodGuess), a.ID())
```
jzzheng22, line 284: Up to you here, but I would use the ConditionType and RequestType, not the numbers

jzzheng22 (approved):
Have a few suggestions but generally they're up to you.

Will be up for a little longer if you do decide to change those.

Would prefer if you wait for a second opinion before merging

pkg/agents/team6/behaviours.go
```diff
@@ -2,48 +2,60 @@ package team6
...
 	// Find new direction required to reach new behaviour prediction
+	// Unconstrained new behaviour - current behaviour
 	updateDir := behaviour(behaviourPrediction)*aConf.maxBehaviourThreshold - a.currBehaviour
+
+	// Apply constraints
 	// Scale movement by stubbornness (minStubborn, maxStubborn) -> (fullMovement, 0)
 	updateMag := updateDir * behaviour(1-aConf.stubbornness)
 	newBehaviour := a.currBehaviour + updateMag
 
+	if newBehaviour.string() != a.currBehaviour.string() {
```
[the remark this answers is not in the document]
MattScottEEE: Yes - string locks to 1 of 4 options in a range, we want to check the behaviour change, not just the numerical change

pkg/agents/team6/treaties.go
mdubied: We'll leave it

pkg/agents/team6/treaties.go
MattScottEEE: Agree fully - makes the code secure. Changing all

pkg/agents/team6/treaties.go
mdubied: done

pkg/agents/team6/treaties.go
MattScottEEE: nice spot - done :)

mdubied: good catch! we changed it

jzzheng22 (approved):
Thanks all :)

MattScottEEE:
Will wait for Jaafar's review in the morning before merging, in case of any last min requests. Other than this, thanks everyone fro the great reviews <3

PR #290, opened 2022-01-10
Add average age on death per agent graph

daryllimyt:
Self explanatory. I could probably reuse some of the other code. to get averages instead of adding a whole new function - lmk if you think this should be done

image


JaafarRammal (approved):
The stats file is getting too big. I propose we split different stats into different components and on the frontend we add titles like "Death Stats" and "Food Stats" (no need for collapsable). Something like this:

---

_Show config button_

---

_Show story button / story status_

---

Death Stats
_Deaths graphs and cards_

Food stats
_Food graphs and cards_

Message state
_Messages graphs and cards_

PR #295, opened 2022-01-11
[Team 6] Fix typo

mdubied:
# Summary
Correct a small mistake in our code I just discovered. 

## Additional Information
As this PR is after the deadline, I leave you the choice of accepting it or not. The correction is rather small and will not affect a lot our agent's strategy. However, the previous version of the code was leading to an agent which was not honest in some cases (if a NewRequestTakeFood message (and not treaty as not implemented) was sent).


MattScottEEE (approved):
This is true - I suppose we should wait for infra to approve as well as me though

jzzheng22 (commented):
Thanks for this, but ultimately will be down to @Silvertrousers to decide

Silvertrousers (approved):
I'm happy to allow this, we haven't started proper testing quite yet (I have an econ exam to revise for) so no harm done

PR #297, opened 2022-01-11
Improve utility calculation

mdubied:
# Summary
Implement an improved version of utility (resolves #270)

## Additional Information
@JaafarRammal the global calculations that should be done on frontend are:
1) utility by agent team over time : sum over agents of one team (utility) / # team agents
2) global utility over time: sum over all agent teams (utility) / # agent teams

### DIFF
--- a/pkg/infra/baseAgent.go
+++ b/pkg/infra/baseAgent.go
@@ -55,7 +55,7 @@ type Base struct {
 	totalFoodSeen  food.FoodType
 	totalHPGained  int
 	totalHPLost    int
-	currentUtility float64
+	utility        float64
 }
 
 func NewBaseAgent(world world.World, agentType agent.AgentType, agentHP int, agentFloor int, id uuid.UUID) (*Base, error) {
@@ -83,7 +83,7 @@ func NewBaseAgent(world world.World, agentType agent.AgentType, agentHP int, age
 		totalFoodSeen:  0,
 		totalHPGained:  0,
 		totalHPLost:    0,
-		currentUtility: 0, //discuss initialization
+		utility:        0, //discuss initialization
 	}, nil
 }
 
@@ -232,7 +232,7 @@ func (a *Base) TakeFood(amountOfFood food.FoodType) (food.FoodType, error) {
 		return 0, &NegFoodError{}
 	}
 	foodTaken := food.FoodType(utilFunctions.MinInt(int(a.tower.currPlatFood), int(amountOfFood), int(a.tower.healthInfo.MaxFoodIntake)))
-	a.utility(amountOfFood, foodTaken)
+	a.updateUtility(amountOfFood, foodTaken)
 	a.updateHP(foodTaken)
 	a.tower.currPlatFood -= foodTaken
 	a.setHasEaten(foodTaken > 0)
@@ -266,23 +266,23 @@ func (a *Base) UpdateHPChange(change int) {
 }
 
 // Updates the utility of a given agent:
-// We take N agents i ∈{1,...,N}that perform the following actions in each iterated round t ∈{1,...,∞}:
-// 1.) Determines the resources it has available, gi ∈[0,1]
-// 2.) Determines its need for resources, qi ∈[0,1]
-// 3.) Makes a provision of resources, pi ∈[0,1]
-// 4.) Makes a demand for resources, di ∈[0,1]
-// 5.) Receives an allocation of resources, ri ∈[0,1]
-// 6.) Makes an appropriation of resources, r′i ∈[0,1]
+func (a *Base) updateUtility(foodRequested, foodTaken food.FoodType) {
+	// We take N agents i ∈{1,...,N}that perform the following actions in each iterated round t ∈{1,...,∞}:
+	// 1.) Determines the resources it has available, gi ∈[0,1]
+	// 2.) Determines its need for resources, qi ∈[0,1]
+	// 3.) Makes a provision of resources, pi ∈[0,1]
+	// 4.) Makes a demand for resources, di ∈[0,1]
+	// 5.) Receives an allocation of resources, ri ∈[0,1]
+	// 6.) Makes an appropriation of resources, r′i ∈[0,1]
 
-// The total resources accrued at the end of a round is hence given by:
-// Ri = r′i + (gi −pi)
-// Which provides the utility of an agent, ui, as:
-// ui = αi(qi) + βi(Ri −qi), if Ri ≥qi
-// 		αi(Ri) −γi(qi −Ri), otherwise
-func (a *Base) utility(foodRequested, foodTaken food.FoodType) {
+	// The total resources accrued at the end of a round is hence given by:
+	// Ri = r′i + (gi −pi)
+	// Which provides the utility of an agent, ui, as:
+	// ui = αi(qi) + βi(Ri −qi), if Ri ≥qi
+	// 		αi(Ri) −γi(qi −Ri), otherwise
+
 	// Hyperparameters that are chosen to make utility commensurate for all agents
-	// These initialisations are chosen by Pitt, but may require tuning
-	// Requires that alpha > gamma > beta;
+	// Requires that alpha > gamma > beta for diminishing returns + greed
 
 	// The weighting assigned to the resources exactly needed to survive
 	alpha := 0.2
@@ -313,9 +313,9 @@ func (a *Base) utility(foodRequested, foodTaken food.FoodType) {
 	R := r_prime + (g - p)
 
 	if R >= q {
-		a.currentUtility = alpha*q + beta*(R-q)
+		a.utility = alpha*q + beta*(R-q)
 	} else {
-		a.currentUtility = alpha*R - gamma*(q-R)
+		a.utility = alpha*R - gamma*(q-R)
 	}
 }
 
@@ -330,7 +330,7 @@ func (a *Base) storyState() logging.AgentState {
 		Floor:     a.floor,
 		Age:       a.age,
 		Custom:    "",
-		Utility:   a.currentUtility,
+		Utility:   a.utility,
 	}
 }
 
--- a/pkg/infra/baseAgent.go
+++ b/pkg/infra/baseAgent.go
@@ -83,7 +83,7 @@ func NewBaseAgent(world world.World, agentType agent.AgentType, agentHP int, age
 		totalFoodSeen:  0,
 		totalHPGained:  0,
 		totalHPLost:    0,
-		utility:        0, //discuss initialization
+		utility:        0,
 	}, nil
 }
 
@@ -288,7 +288,7 @@ func (a *Base) updateUtility(foodRequested, foodTaken food.FoodType) {
 	foodProvision := 0.0
 
 	// 4.) Makes an appropriation of resources, r′i ∈[0,1]
-	// Appropriation is the amount of food an agent actually takes, normalised by the max possible
+	// Amount of food taken, normalised by the max possible intake
 	foodAppropriated := float64(foodTaken) / float64(a.tower.healthInfo.MaxFoodIntake)
 
 	// The total resources accrued at the end of a round is hence given by:
--- a/pkg/infra/baseAgent.go
+++ b/pkg/infra/baseAgent.go
@@ -282,21 +282,23 @@ func (a *Base) UpdateHPChange(change int) {
 func (a *Base) utility(foodRequested, foodTaken food.FoodType) {
 	// Hyperparameters that are chosen to make utility commensurate for all agents
 	// These initialisations are chosen by Pitt, but may require tuning
-	// Requires that alpha > gamma > beta
+	// Requires that alpha > gamma > beta;
+
+	// The weighting assigned to the resources exactly needed to survive
 	alpha := 0.2
+	// The weighting assigned to bonus resources that aren't needed
 	beta := 0.1
+	// The weighting assigned to not getting the resources that are needed
 	gamma := 0.18
 
 	// Available resources normalised by total possible available resources
 	g := float64(a.CurrPlatFood()) / float64(a.tower.maxPlatFood)
 
+	// If an agent is not about to die, it effectively needs 0 food
+	q := 0.0
 	// If the agent is on the verge of dying, food is needed
-	var q float64
 	if a.daysAtCritical == a.HealthInfo().MaxDayCritical {
 		q = 1.0
-		// If an agent is not about to die, it effectively needs 0 food
-	} else {
-		q = 0.0
 	}
 	// No resources are recontributed to the common pool (provision = 0)
 	p := 0.0
--- a/pkg/infra/baseAgent.go
+++ b/pkg/infra/baseAgent.go
@@ -309,7 +309,7 @@ func (a *Base) updateUtility(foodRequested, foodTaken food.FoodType) {
 	// Note that r isn't included. This is because it is just a metric for assessing Ostrom Institution Theory
 	// r >= r_prime (for conventional institution)
 
-	// Resource generation by agent per day is what's available, minus what they give back, plus what they actually take
+	// Resource generation by agent per day is what they actually take, plus the resources available, minus the provisions made
 	R := r_prime + (g - p)
 
 	if R >= q {
--- a/pkg/infra/baseAgent.go
+++ b/pkg/infra/baseAgent.go
@@ -265,35 +265,49 @@ func (a *Base) UpdateHPChange(change int) {
 	}
 }
 
-// Updates the utility of a given agent
-// For this paper, we take N agents i ∈{1,...,N}that perform the following actions in each iterated round t ∈{1,...,∞}:
-// Determines the resources it has available, gi ∈[0,1]
-// Determines its need for resources, qi ∈[0,1]
-// Makes a provision of resources, pi ∈[0,1]
-// Makes a demand for resources, di ∈[0,1]
-// Receives an allocation of resources, ri ∈[0,1]
-// Makes an appropriation of resources, r′i ∈[0,1]
+// Updates the utility of a given agent:
+// We take N agents i ∈{1,...,N}that perform the following actions in each iterated round t ∈{1,...,∞}:
+// 1.) Determines the resources it has available, gi ∈[0,1]
+// 2.) Determines its need for resources, qi ∈[0,1]
+// 3.) Makes a provision of resources, pi ∈[0,1]
+// 4.) Makes a demand for resources, di ∈[0,1]
+// 5.) Receives an allocation of resources, ri ∈[0,1]
+// 6.0 Makes an appropriation of resources, r′i ∈[0,1]
+
 // The total resources accrued at the end of a round is hence given by:
-// Ri = r′i + (gi −pi) (1)
+// Ri = r′i + (gi −pi)
 // Which provides the utility of an agent, ui, as:
-// ui =
-// αi(qi) + βi(Ri −qi), if Ri ≥qi
-// αi(Ri) −γi(qi −Ri), otherwise
+// ui = αi(qi) + βi(Ri −qi), if Ri ≥qi
+// 		αi(Ri) −γi(qi −Ri), otherwise
 func (a *Base) utility(foodRequested, foodTaken food.FoodType) {
+	// Hyperparameters that are chosen to make utility commensurate for all agents
+	// These initialisations are chosen by Pitt, but may require tuning
+	// Requires that alpha > gamma > beta
 	alpha := 0.2
 	beta := 0.1
 	gamma := 0.18
 
+	// Available resources normalised by total possible available resources
 	g := float64(a.CurrPlatFood()) / float64(a.tower.maxPlatFood)
+
+	// If the agent is on the verge of dying, food is needed
 	var q float64
 	if a.daysAtCritical == a.HealthInfo().MaxDayCritical {
 		q = 1.0
+		// If an agent is not about to die, it effectively needs 0 food
 	} else {
 		q = 0.0
 	}
+	// No resources are recontributed to the common pool (provision = 0)
 	p := 0.0
+
+	// Appropriation is the amount of food an agent actually takes, normalised by the max possible
 	r_prime := float64(foodTaken) / float64(a.tower.healthInfo.MaxFoodIntake)
 
+	// Note that r isn't included. This is because it is just a metric for assessing Ostrom Institution Theory
+	// r >= r_prime (for conventional institution)
+
+	// Resource generation by agent per day is what's available, minus what they give back, plus what they actually take
 	R := r_prime + (g - p)
 
 	if R >= q {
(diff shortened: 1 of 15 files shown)


pkg/infra/baseAgent.go
```diff
@@ -262,75 +265,42 @@ func (a *Base) UpdateHPChange(change int) {
...
+	beta := 0.1
+	gamma := 0.18
+
+	g := float64(a.CurrPlatFood()) / float64(a.tower.maxPlatFood)
+	var q float64
+	if a.daysAtCritical == a.HealthInfo().MaxDayCritical {
+		q = 1.0
+	} else {
+		q = 0.0
 	}
```
jzzheng22:
```suggestion
	resNeed := 0.0
	if a.daysAtCritical == a.HealthInfo().MaxDayCritical {
		resNeed = 1.0
	}
```

pkg/infra/baseAgent.go
```diff
@@ -262,75 +265,42 @@ func (a *Base) UpdateHPChange(change int) {
...
-	return float64(intake) / float64(seen)
-}
+	p := 0.0
+	r_prime := float64(foodTaken) / float64(a.tower.healthInfo.MaxFoodIntake)
```
jzzheng22, line 305:
```suggestion
	resApprop := float64(foodTaken) / float64(a.tower.healthInfo.MaxFoodIntake)
```

pkg/infra/baseAgent.go
```diff
@@ -262,75 +265,42 @@ func (a *Base) UpdateHPChange(change int) {
...
+// Which provides the utility of an agent, ui, as:
+// ui =
+// αi(qi) + βi(Ri −qi), if Ri ≥qi
+// αi(Ri) −γi(qi −Ri), otherwise
+func (a *Base) utility(foodRequested, foodTaken food.FoodType) {
+	alpha := 0.2
+	beta := 0.1
+	gamma := 0.18
+
+	g := float64(a.CurrPlatFood()) / float64(a.tower.maxPlatFood)
```
jzzheng22, line 291:
```suggestion
	resAvail := float64(a.CurrPlatFood()) / float64(a.tower.maxPlatFood)
```

pkg/infra/baseAgent.go
```diff
@@ -262,75 +265,56 @@ func (a *Base) UpdateHPChange(change int) {
...
+
+	// If the agent is on the verge of dying, food is needed
+	var q float64
+	if a.daysAtCritical == a.HealthInfo().MaxDayCritical {
+		q = 1.0
+		// If an agent is not about to die, it effectively needs 0 food
+	} else {
+		q = 0.0
+	}
```
jzzheng22, line 300:
```suggestion
	// If an agent is not about to die, it effectively needs 0 food
	resNeed := 0.0
	if a.daysAtCritical == a.HealthInfo().MaxDayCritical {
		resNeed = 1.0
	}
```

pkg/infra/baseAgent.go
```diff
@@ -262,75 +265,56 @@ func (a *Base) UpdateHPChange(change int) {
...
+	// No resources are recontributed to the common pool (provision = 0)
+	p := 0.0
```
jzzheng22, line 302:
```suggestion
	provision := 0.0
```

pkg/infra/baseAgent.go
```diff
@@ -262,75 +265,56 @@ func (a *Base) UpdateHPChange(change int) {
...
-	return simpleQuadraticUtility(hpGainFactor * foodIntakeFactor)
-}
+// Updates the utility of a given agent:
+// We take N agents i ∈{1,...,N}that perform the following actions in each iterated round t ∈{1,...,∞}:
+// 1.) Determines the resources it has available, gi ∈[0,1]
+// 2.) Determines its need for resources, qi ∈[0,1]
+// 3.) Makes a provision of resources, pi ∈[0,1]
+// 4.) Makes a demand for resources, di ∈[0,1]
+// 5.) Receives an allocation of resources, ri ∈[0,1]
+// 6.0 Makes an appropriation of resources, r′i ∈[0,1]
```
MattScottEEE, line 275:
```suggestion
// 6.) Makes an appropriation of resources, r′i ∈[0,1]
```
Typo

pkg/infra/baseAgent.go
```diff
@@ -262,75 +265,56 @@ func (a *Base) UpdateHPChange(change int) {
...
-}
+	// Note that r isn't included. This is because it is just a metric for assessing Ostrom Institution Theory
+	// r >= r_prime (for conventional institution)
 
-// Quadratic function standard form: a(x-h)^2+k
-func quadratic(x, a, h, k float64) float64 {
-	return a*math.Pow((x-h), 2) + k
-}
+	// Resource generation by agent per day is what's available, minus what they give back, plus what they actually take
+	R := r_prime + (g - p)
```
jzzheng22, line 311:
```suggestion
	resAccrued := resApprop + (resAvail - prov)
```

jzzheng22 (changes requested):

pkg/infra/baseAgent.go
```diff
@@ -262,75 +265,56 @@ func (a *Base) UpdateHPChange(change int) {
...
-*/
-func foodIntakeFactor(intake, seen food.FoodType) float64 {
-	// If an agent hasn't seen any food, has 0 utility
-	if seen == 0 {
-		return 0
-	}
-	return float64(intake) / float64(seen)
-}
+	// Note that r isn't included. This is because it is just a metric for assessing Ostrom Institution Theory
+	// r >= r_prime (for conventional institution)
```
jzzheng22, line 308: Irrelevant here. Put it in the report if you want.

pkg/infra/baseAgent.go
```diff
@@ -262,75 +265,56 @@ func (a *Base) UpdateHPChange(change int) {
...
 
-// Quadratic function standard form: a(x-h)^2+k
-func quadratic(x, a, h, k float64) float64 {
-	return a*math.Pow((x-h), 2) + k
-}
+	// Resource generation by agent per day is what's available, minus what they give back, plus what they actually take
```
jzzheng22, line 310: Would order this so it is in the same order as the formula.

jzzheng22 (changes requested):

pkg/infra/baseAgent.go
MattScottEEE: I think it's important. It's a variable that's included inside of the function description that magically doesn't appear. Having this comment explicitly shows that it's a utility calculation, not a validity calculation

pkg/infra/baseAgent.go
```diff
@@ -262,75 +265,56 @@ func (a *Base) UpdateHPChange(change int) {
...
+// 6.) Makes an appropriation of resources, r′i ∈[0,1]
+
+// The total resources accrued at the end of a round is hence given by:
+// Ri = r′i + (gi −pi)
+// Which provides the utility of an agent, ui, as:
+// ui = αi(qi) + βi(Ri −qi), if Ri ≥qi
+// 		αi(Ri) −γi(qi −Ri), otherwise
+func (a *Base) utility(foodRequested, foodTaken food.FoodType) {
+	// Hyperparameters that are chosen to make utility commensurate for all agents
+	// These initialisations are chosen by Pitt, but may require tuning
```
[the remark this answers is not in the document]
MattScottEEE: Fine to delete, but should be made known to the cohort

pkg/infra/baseAgent.go
MattScottEEE: Fair.

Since this formula comes direct from literature (and is explicitly defined above), I would rather keep the variable names as letters

pkg/infra/baseAgent.go
jzzheng22: I wouldn't put it in the function description then.

pkg/infra/baseAgent.go
```diff
@@ -262,75 +265,56 @@ func (a *Base) UpdateHPChange(change int) {
...
+// 4.) Makes a demand for resources, di ∈[0,1]
+// 5.) Receives an allocation of resources, ri ∈[0,1]
+// 6.) Makes an appropriation of resources, r′i ∈[0,1]
+
+// The total resources accrued at the end of a round is hence given by:
+// Ri = r′i + (gi −pi)
+// Which provides the utility of an agent, ui, as:
+// ui = αi(qi) + βi(Ri −qi), if Ri ≥qi
+// 		αi(Ri) −γi(qi −Ri), otherwise
+func (a *Base) utility(foodRequested, foodTaken food.FoodType) {
```
JaafarRammal, line 282:
```suggestion
func (a *Base) updateUtility(foodRequested, foodTaken food.FoodType) {
```

pkg/infra/baseAgent.go
jzzheng22: It's more readable this way - report can use the letters, but here it's quite obscure.

pkg/infra/baseAgent.go
```diff
@@ -262,75 +265,56 @@ func (a *Base) UpdateHPChange(change int) {
...
-}
+	// Resource generation by agent per day is what's available, minus what they give back, plus what they actually take
+	R := r_prime + (g - p)
 
-// Utility function is a strictly increasing function that starts at 0,0 ends at 1,1
-func simpleQuadraticUtility(x float64) float64 {
-	// -(x-1)^2 + 1
-	return utilFunctions.RestrictToRange(0, 1, quadratic(x, -1, 1, 1))
+	if R >= q {
+		a.currentUtility = alpha*q + beta*(R-q)
```
JaafarRammal, line 314:
```suggestion
		a.utility = alpha*q + beta*(R-q)
```

pkg/infra/baseAgent.go
```diff
@@ -262,75 +265,56 @@ func (a *Base) UpdateHPChange(change int) {
...
+	} else {
+		a.currentUtility = alpha*R - gamma*(q-R)
```
JaafarRammal, line 316:
```suggestion
		a.utility = alpha*R - gamma*(q-R)
```

pkg/infra/baseAgent.go
```diff
@@ -344,7 +328,7 @@ func (a *Base) storyState() logging.AgentState {
 		Floor:     a.floor,
 		Age:       a.age,
 		Custom:    "",
-		Utility:   a.utility(),
+		Utility:   a.currentUtility,
```
JaafarRammal, line 331:
```suggestion
		Utility:   a.utility,
```

pkg/infra/tower.go
```diff
@@ -97,14 +97,15 @@ func (t *Tower) Reshuffle() {
...
 
 func (t *Tower) endOfDay() {
-	t.dayInfo.CurrDay += 1
 	for _, agent := range t.Agents {
 		agent := agent.BaseAgent()
 		agent.hpDecay(t.healthInfo)
 		agent.increaseAge()
 		agent.updateTreaties()
-		t.stateLog.LogUtility(t.dayInfo, agent.agentType, agent.utility(), agent.IsAlive())
+		t.stateLog.LogUtility(t.dayInfo, agent.agentType, agent.currentUtility, agent.IsAlive())
```
JaafarRammal, line 105:
```suggestion
		t.stateLog.LogUtility(t.dayInfo, agent.agentType, agent.utility, agent.IsAlive())
```

pkg/infra/tower.go
```diff
@@ -97,14 +97,15 @@ func (t *Tower) Reshuffle() {
...
+		agent.currentUtility = 0
```
JaafarRammal, line 106:
```suggestion
		agent.utility = 0
```
Can have an `EndOfDay` function in agents and we move all of these state updates to it, although can be a different PR

pkg/infra/baseAgent.go
mdubied: If we use specific variable names in the report, we should probably use the same ones in the code if possible (especially if the formulas are taken from literature). I would rather keep it as it is

JaafarRammal: I agree with Jason. Formulas define the concept, but it's better to set readable variable names in the code so the logic can be easily evaluated. E.g: if `q` represents allocation in the formula, it is easier to use a variable name like `foodAlloc` rather than `q`, so when someone is reviewing the code, they can quickly go "aha,`foodAlloc` is being computed correctly" or "hmm I think `foodAlloc` can be obtained in a better way"

pkg/infra/tower.go
MattScottEEE: Would be nice with a separate PR if ok

pkg/infra/baseAgent.go
```diff
@@ -262,75 +265,58 @@ func (a *Base) UpdateHPChange(change int) {
 	}
 }
 
-func (a *Base) utility() float64 {
-	// TODO: Improve utility calculation
-	// Currently just use ratio between food taken and food seen
-	// Somehow also factor
-	// Ideas: piecewise/bounded function with log, concave quadratic
+// Updates the utility of a given agent:
```
jzzheng22, line 268:
```suggestion
```

pkg/infra/baseAgent.go
```diff
@@ -262,75 +265,58 @@ func (a *Base) UpdateHPChange(change int) {
...
+func (a *Base) updateUtility(foodRequested, foodTaken food.FoodType) {
+	// We take N agents i ∈{1,...,N}that perform the following actions in each iterated round t ∈{1,...,∞}:
```
jzzheng22, line 270:
```suggestion
```

pkg/infra/baseAgent.go
```diff
@@ -262,75 +265,58 @@ func (a *Base) UpdateHPChange(change int) {
...
+	// 1.) Determines the resources it has available, gi ∈[0,1]
+	// 2.) Determines its need for resources, qi ∈[0,1]
+	// 3.) Makes a provision of resources, pi ∈[0,1]
+	// 4.) Makes a demand for resources, di ∈[0,1]
+	// 5.) Receives an allocation of resources, ri ∈[0,1]
+	// 6.) Makes an appropriation of resources, r′i ∈[0,1]
```
jzzheng22, line 276: I mean put each line above the stage where it is relevant, so move the comment 1.) to above the line where `resAvail` is declared, and so on.

Don't need comment 5.

pkg/infra/baseAgent.go
```diff
@@ -262,75 +265,58 @@ func (a *Base) UpdateHPChange(change int) {
...
+	beta := 0.1
+	// The weighting assigned to not getting the resources that are needed
+	gamma := 0.18
+
+	// Available resources normalised by total possible available resources
+	g := float64(a.CurrPlatFood()) / float64(a.tower.maxPlatFood)
+
+	// If an agent is not about to die, it effectively needs 0 food
+	q := 0.0
+	// If the agent is on the verge of dying, food is needed
```
jzzheng22, line 299:
```suggestion
	// If an agent is not about to die, it effectively needs 0 food
	// If the agent is on the verge of dying, food is needed
	q := 0.0
```

pkg/infra/baseAgent.go
```diff
@@ -262,75 +265,58 @@ func (a *Base) UpdateHPChange(change int) {
...
+	if a.daysAtCritical == a.HealthInfo().MaxDayCritical {
+		q = 1.0
+	}
+	// No resources are recontributed to the common pool (provision = 0)
```
jzzheng22, line 303:
```suggestion
	}
	
	// No resources are recontributed to the common pool (provision = 0)
```

jzzheng22 (changes requested):

pkg/infra/baseAgent.go
MattScottEEE: We can do it this way :)

I'll push with better variables

pkg/infra/baseAgent.go
```diff
@@ -82,6 +83,7 @@ func NewBaseAgent(world world.World, agentType agent.AgentType, agentHP int, age
 		totalFoodSeen:  0,
 		totalHPGained:  0,
 		totalHPLost:    0,
+		utility:        0, //discuss initialization
```
jzzheng22, line 86:
```suggestion
		utility:        0,
```

pkg/infra/baseAgent.go
```diff
@@ -262,75 +265,57 @@ func (a *Base) UpdateHPChange(change int) {
...
+	
+	// No resources are recontributed to the common pool (provision = 0)
+	foodProvision := 0.0
 
-	hpGainFactor := netHPGainFactor(a.totalHPGained, a.totalHPLost, a.tower.healthInfo.MaxHP)
-	foodIntakeFactor := foodIntakeFactor(a.totalFoodTaken, a.totalFoodSeen)
-	// this means the multiplcation of these values can be larger than 1
-	return simpleQuadraticUtility(hpGainFactor * foodIntakeFactor)
-}
+	// Appropriation is the amount of food an agent actually takes, normalised by the max possible
```
jzzheng22, line 305:
```suggestion
	// Amount of food taken, normalised by the max possible intake
```

pkg/infra/baseAgent.go
MattScottEEE: This refers to the if case, hence q = 1

pkg/infra/baseAgent.go
```diff
@@ -262,75 +265,57 @@ func (a *Base) UpdateHPChange(change int) {
...
-Net HP Gain Factor.
-This one uses net hp gain as opposed to hp gain/loss ratio.
-*/
-func netHPGainFactor(gain, loss, maxhp int) float64 {
-	netHPGain := gain - loss
-	multiplier := 1 + float64(netHPGain)/float64(maxhp) // reward positive net hp gain, punish otherwise
-	return utilFunctions.RestrictToRange(0, math.Inf(1), multiplier)
-}
+	// Note that r isn't included. This is because it is just a metric for assessing Ostrom Institution Theory
+	// r >= r_prime (for conventional institution)
```
jzzheng22, line 309:
```suggestion
```

jzzheng22 (dismissed):

pkg/infra/baseAgent.go
```diff
@@ -262,75 +265,57 @@ func (a *Base) UpdateHPChange(change int) {
...
+	// Available resources normalised by total possible available resources
+	foodAvailable := float64(a.CurrPlatFood()) / float64(a.tower.maxPlatFood)
+
+	// If an agent is not about to die, it effectively needs 0 food
+	foodToSurvive := 0.0
+	// If the agent is on the verge of dying, food is needed
+	if a.daysAtCritical == a.HealthInfo().MaxDayCritical {
+		foodToSurvive = 1.0
+	}
+
```
woonmoon, line 301: Is this strictly the case? Surely when it's just critical and not just on its dying day it can count as needing food?

MattScottEEE: But you can still survive. The only time you literally need food is when you're critical and about to die

MattScottEEE: Unless I'm mistaken - can you be critical and die, even if it's not the dying day?

mdubied: I would say it is only "needed" the very last day. This is also the way to match the specs, i.e., having agents oscillating between critical state and WeakLevel while spending more time in the critical state.

As a comment, note also that 1.0 is not really the food to survive, but rather ressourceNeeded/maxRessourceNeeded, so that we have q_i in [0,1]. The actual amount of food to survive is actually 2 with the current health parameters

MattScottEEE: Hm, you might be right on this one actually - any food here would be good utility

MattScottEEE: @mdubied I'm happy to go with Gowoon on this one?

MattScottEEE: > I would say it is only "needed" the very last day. This is also the way to match the specs, i.e., having agents oscillating between critical state and WeakLevel while spending more time in the critical state.
> 
> The variable name foodToSurvive is actually not perfect (but ok I think)
> 
> 1.0 is not really the food to survive, but rather ressourceNeeded/maxRessourceNeeded, so that we have q_i in [0,1]. The actual amount of food to survive is actually 2 with the current health parameters

My only issue with this (that makes me want to go with Gowoon) is that taking 2 food while in critical would otherwise be seen as wasted utility. I think this is probably unfair, even though they can technically survive a bit longer. Agents shouldn't be penalised for returning to weak when in critical

mdubied: @MattScottEEE you can not die if it's not "the dying day" (what a sentence hahah)

I am not completely sure about it. I completely agree that the food at critical is good utility. But it we push the reasoning, this same amount of food is also good when we are at WeakLevel. With the same amount of food you can either go from ciritical to weak, or stay at weak if you already at weak. Therefore, I would say that the food is really "needed" only to avoid dying, i.e., when `DaysAtCritical=MaxDayCritical`

woonmoon: If you're on day 2 at a critical state, you skip eating because day 3 is the last possible day and the platform gets to you on the next day (day 3) and it's empty (probably not an unusual case) - you're dead, you should have eaten on day 2 and you needed that food when it got to you.
This isn't exact because there are no guarantees but I think the critical cut-off is a good place to put the threshold and a useful distinction against idk `weak` for example.

MattScottEEE: Hm, overall I think that we should be rewarding agents for taking food at the critical stage

How about a compromise:

Taking food when at the very final day rewards a q=1.0
Taking food when just at critical (before final day) rewards 0.5<q<1.0

@woonmoon @mdubied

MattScottEEE: > If you're on day 2 at a critical state, you skip eating because day 3 is the last possible day and the platform gets to you on the next day (day 3) and it's empty (probably not an unusual case) - you're dead, you should have eaten on day 2 and you needed that food when it got to you.
> This isn't exact because there are no guarantees but I think the critical cut-off is a good place to put the threshold and a useful distinction against idk `weak` for example.

Whilst you may die from this, that doesn't really matter, as you'd technically be getting more utility for your food if you waited another day. It's just unlucky. At the same time, you definitely should get some utility for even waiting until critical. We could parameterise q as a mapping from days critical to 0->1 (daysCrit / maxDaysCrit?)

mdubied: It is really a good question to answer @woonmoon !
The argument you givis not completely satisfying in my opinion. Using the same reasoning, I could say that you need to eat already even earlier, for example when HP=20 to avoid the risky scenario where you get nothing on day 3 of being at critical. But we agree that eating when HP = 20 is not "needed"
 I quite like your compromise @MattScottEEE though

woonmoon: @mdubied Yes but isn't this why we categorized HP into "how desperately do you need to eat"? Yes you could be unlucky and the platform could arrive empty at you for 20 days in a row and you're just dead. It's just not good utility for an agent to eat when they're healthy but the same can't be said if you know you're ~3 days out from dying. I agree with @MattScottEEE that we should not be penalizing agents for eating when they're critical.

JaafarRammal (approved):
LGTM and tested it
FSA, utility log size increases linearly with number of agents and number of days (each agent gets its utility logged once per day)

JaafarRammal (approved):

daryllimyt (approved):
LGTM Thanks for this

woonmoon (approved):
Thanks for making the change guys.

PR #298, opened 2022-01-11
[HOTFIX] Timeout

JaafarRammal:
Move timeout call to correct place


jzzheng22 (approved):
Does Victor need to check it too or?

JaafarRammal:
DMed victor all good!

PR #300, opened 2022-01-12
[Team 6] Send replies when treaties are rejected

mdubied:
# Summary
When running simulations, we found out that we were not sending negative replies when rejecting treaties. This is now fixed by just adding a reply message.

As the PR is after the deadline of Monday, you can reject it. However, the logs concerning treaties (rejection) will be wrong for our team if not accepted


jzzheng22 (changes requested):
The `requestID` field is the ID of the message you are responding to, not the ID of the agent that proposed the message.

This needs to be fixed in both the acceptance and rejection cases.

I believe that if you don't respond, it's assumed to be a rejection, but @daryllimyt  can confirm that.

I'll leave whether to accept this change up to @Silvertrousers, though I suspect it's fine.

mdubied:
I'll change the `requestID` then! 
Concerning "I believe that if you don't respond, it's assumed to be a rejection", I don't think so, as we had no treaty rejection showing in the log before making the change described in this PR

pkg/agents/team6/communication.go
```diff
@@ -254,6 +254,8 @@ func (a *CustomAgent6) HandleProposeTreaty(msg messages.ProposeTreatyMessage) {
 		a.Log("I accepted a treaty proposed from ", infra.Fields{"senderFloor": msg.SenderFloor(), "myFloor": a.Floor(), "my social motive": a.currBehaviour.string()})
 
 	} else {
+		reply := messages.NewTreatyResponseMessage(a.ID(), a.Floor(), msg.SenderFloor(), false, treaty.ID(), treaty.ProposerID())
+		a.SendMessage(reply)
 		a.updateTrust(-1, msg.SenderID()) // bad treaty - these guys are trying to sabotage us >:)
```
woonmoon, line 259: Why would you not use `proposeTreatyMessage.Reply(...)`?

woonmoon (commented):
I agree with Jason's review but also this could have been avoided using the `message.Reply(agent.ID(), agent.Floor(), msg.SenderFloor(), response)`

pkg/agents/team6/communication.go
mdubied: Good spot! We did not know about the existence of the `proposeTreatyMessage.Reply()` function. We just used the less direct way by creating a new `NewTreatyResponseMessage` by ourself

mdubied:
@jzzheng22 @woonmoon I adressed both of your comment by using the `.Reply()` function that was proposed. I hope it looks good now :)

jzzheng22 (commented):
Thanks. Please wait for review and approval from Cyrus.

daryllimyt:
> The `requestID` field is the ID of the message you are responding to, not the ID of the agent that proposed the message.
> 
> This needs to be fixed in both the acceptance and rejection cases.
> 
> I believe that if you don't respond, it's assumed to be a rejection, but @daryllimyt can confirm that.
> 
> I'll leave whether to accept this change up to @Silvertrousers, though I suspect it's fine.

@jzzheng22 no currently we're calling `StoryLog()` if the message sent is a `treatyResponse`, which is overloaded to produce a yes/no for accept/reject. Note that this is from the agents POV, i.e. accept -> number that this agent has accepted, reject -> number that this agent rejected. Not to be confused with whether an agent's treaties get accepted/rejected

Silvertrousers (approved):

jzzheng22 (approved):

PR #301, opened 2022-01-12
Frontend improvements

JaafarRammal:


jzzheng22 (commented):
Can I be annoying and ask for these changes too <3

1. Default to 'Use Food Per Agent' by default
2. Simulation Length default to 100 days?
3. Reshuffle period default of 7 days?
4. TimeOut longer (say 180? or even longer?)
5. Team1 and RandomAgent default values to 0 
6. Remove ability to change Agents Per Floor and Agents Max HP (and maybe even Initial HP)
7. The Treaties and Messages graphs have columns for agents which weren't in the tower - do we want that?
8. Can we replace the Config bit which shows agents per floor and show shuffle period?

jzzheng22:
Thank you for your work <3

jzzheng22:
Also I'm assuming you're already aware, but commenting this just in case - you can now open a log while the simulation is still running, but you can't open the 'Other Stats' section - that still causes a crash.

JaafarRammal:
Other stats contains "messages" graphs which are only logged at the end of a Sim (it even crashes if there is a timeout). I'll try to resolve the timeout thing, and for the "sim is running" ill just disable the section until the sim finishes?

JaafarRammal:
> 7- The Treaties and Messages graphs have columns for agents which weren't in the tower - do we want that?

@daryllimyt worked on the graphs, I think he's better suited into resolving this one if it's an issue!

jzzheng22:
@JaafarRammal  btw Death log doesn't need to be disabled until sim finishes, that one is fine

JaafarRammal:
Yep reenabled it

jzzheng22 (dismissed):

daryllimyt:
also thanks @JaafarRammal for consolidating the death stats :)

daryllimyt:
I'm gonna do 7 in a different PR so you can get the other stuff in first

jzzheng22 (approved):

daryllimyt (approved):

PR #304, opened 2022-01-13
Update messages and treaties graphs to only show data for present agents

daryllimyt:
Self explanatory. Additionally made util functions return tuples of labels and data.

image


JaafarRammal (changes requested):
I might sound annoying, but would it be possible to pull from `result.config` directly which agents are present in the sim instead of creating it again on the backend :) It's a bit convoluted since the frontend already knows which agents exist and which don't

JaafarRammal (dismissed):

JaafarRammal (approved):

PR #305, opened 2022-01-13
Add scripts to run backend and frontend

jzzheng22:
# Summary


scripts/frontend.sh
```diff
@@ -0,0 +1,5 @@
+set -euxo pipefail
+
+cd frontend
+npm install
```
JaafarRammal, line 4:
```suggestion
npm ci
```

JaafarRammal (changes requested):

scripts/frontend.sh
jzzheng22: npm install runs a lot faster - typically it's fine for most use cases

JaafarRammal: it's just that npm install might modify package.json
TBF we can spot that in PRs, let it be npm install

JaafarRammal (approved):

JaafarRammal (approved):

PR #311, opened 2022-01-13
[Team 7] Update greediness calculation parameter

moinbukhari:
# Summary

Fixed greediness bug


jzzheng22 (commented):
Fine from a code perspective, but up to @Silvertrousers  about whether this should be accepted.

Silvertrousers (approved):

PR #314, opened 2022-01-17
Logging food per floor average

JaafarRammal:
# Summary

image


jzzheng22 (approved):
i dont have time to review this properly so i hope you know what you're doing

is there a reason why `foodFloorMap` is a slice and not a map though?

JaafarRammal:
Hmm good point it should be name foodFloorSlice :)

